PR #1, opened 2025-01-12
First pass at debranding the agent

binford2k:
This will debrand the common things like `puppet help` and the man
pages. I think I got all the command output that includes the name of
the tool.

It also does a quick pass over the docs generation system. We aren't
building these yet, but we will at some point.

If you'd like to help, then the easiest way to do so is to check out
this branch/PR and directly add your own commits.

Guidelines
- Do not change any command names.
- *Do not change anything functional!!*
- If any command output includes the name of the tool, change it as
  appropriate.
    - if the name is extraneous, then just remove it. For example in the
      message "cannot find puppet server" I just removed the word
      "puppet".
- When there is a reference to 'Puppet' that means 'the Puppet language'
  then change it to 'the Puppet language' to be more precise. (OpenVox
  implements the Puppet language. There's no such thing as an OpenVox
  language.)
- Do minor content editing for readability as appropriate.
- We are currently focusing only on user experience and branding. It
  should be very clear to a user that they're running the OpenVox
  implementation of Puppet. It's not yet important that all the internal
  and developer documentation or comments reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,6 @@
-# Puppet
+# OpenVox
 
-![RSpec tests](https://github.com/puppetlabs/puppet/workflows/RSpec%20tests/badge.svg)
-[![Gem Version](https://badge.fury.io/rb/puppet.svg)](https://badge.fury.io/rb/puppet)
-[![Inline docs](https://inch-ci.org/github/puppetlabs/puppet.svg)](https://inch-ci.org/github/puppetlabs/puppet)
-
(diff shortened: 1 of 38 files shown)


binford2k:
I took another pass at this and am up to something like 2k lines of change. I'm not comfortable with this large of a change yet, so I'm going to take another approach. I'm just going to update the `puppet help`, README, and man pages.

PR #4, opened 2025-01-14
First pass at debranding

binford2k:
This should update to OpenVox in help output and man pages. It also adds
a debranded README and gemspec. Supercedes #2, removing generated man
pages by request. We'll generate them after we're done making changes.

If you'd like to help, feel free to check out this branch/PR and push
your own commits back to the branch. Stay within scope though,
we're focusing first on the man pages and any user interaction. 
We'll get to docs later, that's gonna be a beast.

Guidelines
- Do not change any command names.
- *Do not change anything functional!!*
- If any command output includes the name of the tool, change it as
  appropriate or add something like "provided by OpenVox"
    - if the name is extraneous, then just remove it. For example in the
      message "cannot find puppet server" I just removed the word
      "puppet".
- When there is a reference to 'Puppet' that means 'the Puppet language'
  then change it to 'the Puppet language' to be more precise. (OpenVox
  implements the Puppet language. There's no such thing as an OpenVox
  language.)
- Do minor content editing for readability as appropriate.
- We are currently focusing only on user experience and branding. It
  should be very clear to a user that they're running the OpenVox
  implementation of Puppet. It's not yet important that all the internal
  and developer documentation or comments reflect this.

PR #9, opened 2025-01-16
update generated man pages for the debrand

binford2k:
This will probably have more LOC removed than added. The updated
toolchain seems to generate equivalent rendered output with smaller
source files. For example, the old pages had a whole ton of lines that
were empty other than a single `.` and those were removed.

In any case, it's just rendered output. If you'd like to inspect the
pages themselves, then check out this branch and run `man` directly on
the files, such as

```
$ man man/man5/puppet-agent.8
```


genebean (approved):
I really looked at #4 and trust this accurately represents what was there.

PR #18, opened 2025-02-28
remove header in gzipped files

kapouik:
Remove header in man pages when they are compress. It is a step for reproducible build.


kapouik:
https://github.com/OpenVoxProject/puppet/pull/10 <--- duplicate. I close my PR

PR #22, opened 2025-03-04
Prevents spurious clientcert warnings in serverless mode

### LINKED ISSUE #21: [Bug]: puppet apply warns about missing client certificates

### Describe the Bug

When running puppet in masterless mode, where you only run `puppet apply` on a manifest file, rather than a puppet master and puppet agent, the resulting server does not require certificates because there is no master to communicate with.

When running puppet apply however, there is an false positive warning about the client certificate not existing, when it's not relevant to that execution mode.

### Expected Behavior

No warning for a client certificate not existing, when one is not expected to exist.

### Steps to Reproduce

Make a minimal test.pp manifest file:

```pp
file {
  '/tmp/composer':
    mode   => '0755',
    source => 'https://getcomposer.org/download/2.8.6/composer.phar';
}
```

Execute:
```sh
puppet apply test.pp
```

Log:
```
Notice: Compiled catalog for phenom in environment production in 0.01 seconds
Warning: Private key for 'fqdn' does not exist
Warning: Client certificate for 'fqdn' does not exist
Notice: /Stage[main]/Main/File[/tmp/composer]/ensure: defined content as '{mtime}2025-02-25 12:03:50 UTC'
Notice: Applied catalog in 0.25 seconds
```

### Environment

Version: 8.13
Platform: Gentoo Linux

### Additional Context

_No response_

### Relevant log output

```shell

```

binford2k:
When there are no clientcerts, Puppet will warn when it creates an
`SSLContext` for HTTPS operations. This situation occurs when you run
entirely serverless and never generate clientcerts. It's spurious in
that case, so we don't actually need to warn about it.

This behaviour was added in https://github.com/OpenVoxProject/puppet/commit/3f7f830f0aef6fd3d1bf77a441d4000b9a586ed0
so that the new HTTP client could download files via HTTPS from the
puppetserver (for example, the way that pe_repo) works.

To prevent this being a failure when running `puppet apply` in
serverless mode, it explicitly marks the clientcerts as optional in
https://github.com/OpenVoxProject/puppet/blob/06bc441333c640678c9adb26412c6cb923af7f6b/lib/puppet/ssl/ssl_provider.rb#L98
and
https://github.com/OpenVoxProject/puppet/blob/06bc441333c640678c9adb26412c6cb923af7f6b/lib/puppet/ssl/ssl_provider.rb#L103

This goes one step further and sets the output to `INFO` rather than `WARN`
when running `puppet apply`.

This does have one small edge case. If,

1. You intend to run a standard server/agent setup, and
2. Before ever running `puppet agent -t` you run `puppet apply` for
   provisioning purposes, and
3. Part of that Puppet run attempts to download a file from the
   puppetserver

Then you will get a certificate validation error and the HTTPS request
will fail silently with only an `INFO` message as a hint explaining why.

To fix it, you obviously just generate and sign the clientcerts.

I think this is an acceptable tradeoff, but would like other opinions.
This will need specs before merging.

Fixes #21

### DIFF
--- a/lib/puppet/ssl/ssl_provider.rb
+++ b/lib/puppet/ssl/ssl_provider.rb
@@ -97,12 +97,14 @@ class Puppet::SSL::SSLProvider
       cert_provider = Puppet::X509::CertProvider.new
       private_key = cert_provider.load_private_key(Puppet[:certname], required: false)
       unless private_key
-        Puppet.warning("Private key for '#{Puppet[:certname]}' does not exist")
+        msg = "Private key for '#{Puppet[:certname]}' does not exist"
+        Puppet.run_mode.name == :user ? Puppet.info(msg) : Puppet.warning(msg)
       end
 
       client_cert = cert_provider.load_client_cert(Puppet[:certname], required: false)
       unless client_cert
-        Puppet.warning("Client certificate for '#{Puppet[:certname]}' does not exist")
+        msg "Client certificate for '#{Puppet[:certname]}' does not exist"
+        Puppet.run_mode.name == :user ? Puppet.info(msg) : Puppet.warning(msg)
       end
 
       if private_key && client_cert


lib/puppet/ssl/ssl_provider.rb
```diff
@@ -97,12 +97,14 @@ def create_system_context(cacerts:, path: Puppet[:ssl_trust_store], include_clie
...
       unless private_key
-        Puppet.warning("Private key for '#{Puppet[:certname]}' does not exist")
+        msg = "Private key for '#{Puppet[:certname]}' does not exist"
+        Puppet.run_mode.name == :user ? Puppet.info(msg) : Puppet.warning(msg)
       end
 
       client_cert = cert_provider.load_client_cert(Puppet[:certname], required: false)
       unless client_cert
-        Puppet.warning("Client certificate for '#{Puppet[:certname]}' does not exist")
+        msg "Client certificate for '#{Puppet[:certname]}' does not exist"
```
anthonyryan1: Missing the = to make this an assignment, currently trying to call msg()

ffrank:
Seems fine, https://github.com/OpenVoxProject/puppet/commit/3f7f830f0aef6fd3d1bf77a441d4000b9a586ed0 even mentions that a missing client cert is supposed to be ignored.

How silently does a request fail in the described edge case? The resource should fail indicating a 4xx code from the server, no?

binford2k:
@ffrank yes, the connection itself will fail and any code paths that depend on it will fail. Those are often swallowed up by some exception handling. What I meant was that the message describing what actually went wrong would only be logged as `INFO` instead of an error.

binford2k:
I just realized that I left this as draft

nmburgan (approved):

PR #35, opened 2025-04-16
Fix unit tests missed during the de-brandification

seanmil:


adamboutcher:
But surely this isn't branding but referring to the command/binary which is still named puppet...

nmburgan (approved):

PR #37, opened 2025-04-16
Fix unit test CSR versions for OpenSSL 3.4 change

seanmil:
The only valid CSR version is version 1, which is encoded with "csr.version = 0". Versions of OpenSSL prior to 3.4 would allow other versions to be set / versions to be unset but OpenSSL 3.4 is strict about only allowing version 1.

This change updates the unit tests and test fixtures to all specify or contain version 1 CSRs.

### DIFF
--- a/spec/fixtures/ssl/request.pem
+++ b/spec/fixtures/ssl/request.pem
@@ -1,6 +1,6 @@
 Certificate Request:
     Data:
-        Version: 3 (0x2)
+        Version: 1 (0x0)
         Subject: CN=pending
         Subject Public Key Info:
             Public Key Algorithm: rsaEncryption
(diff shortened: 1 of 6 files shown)


spec/fixtures/ssl/request.pem
```diff
@@ -1,10 +1,10 @@
 Certificate Request:
     Data:
-        Version: 3 (0x2)
```
bastelfreak, line 3: Could you please document how you generated the fixtures?

seanmil: > Could you please document how you generated the fixtures?

Yes. I had meant to, and then forgot. Working on it.

seanmil: Okay, I pushed an update with a script which can generate the ones in question.  I wasn't prepared to take on the whole ssl/fixtures directory this time around.

nmburgan (approved):

PR #38, opened 2025-04-18
Fix given_or_derived behavior in Pops object init_hash

seanmil:
In create_init_hash_type the type specification for an init hash is built.  Currently a given_or_derived type is always defined as a required init_hash attribute, which defeats the ability to leave it unspecified and use the derived value defaults.

By defining given_or_derived types always as optional it allows passing a value in the init_hash or leaving it out and using the derived value.

### DIFF
--- a/lib/puppet/pops/types/p_object_type.rb
+++ b/lib/puppet/pops/types/p_object_type.rb
@@ -876,7 +876,7 @@ class PObjectType < PMetaType
     struct_elems = {}
     attributes(true).values.each do |attr|
       unless attr.kind == ATTRIBUTE_KIND_CONSTANT || attr.kind == ATTRIBUTE_KIND_DERIVED
-        if attr.value?
+        if attr.value? || attr.kind == ATTRIBUTE_KIND_GIVEN_OR_DERIVED
           struct_elems[TypeFactory.optional(attr.name)] = attr.type
         else
           struct_elems[attr.name] = attr.type
--- a/spec/unit/pops/types/p_object_type_spec.rb
+++ b/spec/unit/pops/types/p_object_type_spec.rb
@@ -697,18 +697,16 @@ describe 'The Object Type' do
   end
 
   context 'when producing an init_hash_type' do
-    it 'produces a struct of all attributes that are not derived or constant' do
+    it 'produces a struct of all attributes excepting derived or constant' do
       t = parse_object('MyObject', <<-OBJECT)
         attributes => {
           a => { type => Integer },
-          b => { type => Integer, kind => given_or_derived },
-          c => { type => Integer, kind => derived },
-          d => { type => Integer, kind => constant, value => 4 }
+          b => { type => Integer, kind => derived },
+          c => { type => Integer, kind => constant, value => 4 }
         }
       OBJECT
       expect(t.init_hash_type).to eql(factory.struct({
         'a' => factory.integer,
-        'b' => factory.integer
       }))
     end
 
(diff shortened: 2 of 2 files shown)


seanmil:
I spent quite a bit of time reading through the code and Henrik's series of blog posts before I came to the decision this PR was the right direction.  I'd love for @hlindberg to comment as to his thoughts, if he's so inclined.

hlindberg:
Oh dear, it was ages ago since I had all that current in my head.
Could you help me with an example of usage as it at the moment is a bit too abstract to grok - especially what any possible downsides of the change could be.

hlindberg:
Byw, do you have access to the (very long) document about PCore?
If not I will go find it for you.

hlindberg:
Here is the PCore design document, which was the presentation and motivation for the type system and PCore. Much of this was later turned into specs and implementation. I think the document is still a good read for getting the background and ideas behind the type system. The doc is open to anyone with the link: https://docs.google.com/document/d/1Oq4VFISrhU44nnKWsYNB_Q_OLnPnodJz67pB8d53jys/edit?tab=t.0

seanmil:
@hlindberg sure, thanks for responding! Thanks for the link to the PCore doc - I had it, somewhere, but forgot I did. I put together a simple (but dumb) example here: https://github.com/seanmil/example_given_or_derived

Without this patch the given_or_derived argument is treated as required during initialization, which means that a conforming implementation would never be able to use the "derived" part.  Given the PCore document (which is the clearest reference I've found) I believe this is the correct behavior.

Here is the output of the example I linked without the patch:

```
Error: Evaluation Error: Error while evaluating a Method call, The function 'new_Example' was called with arguments it does not accept. It expects one of:
  (Struct[{'int' => Integer, 'maybe_half' => Numeric}] hash)
    rejected: parameter 'hash' expects a value for key 'maybe_half'
  (Integer int, Numeric maybe_half)
    rejected: expects 2 arguments, got 1 (file: modpath/example_given_or_derived/manifests/init.pp, line: 2, column: 21)
```

And here it is with the patch:

```
Notice: Scope(Class[Example_given_or_derived]): Ex1.int: 44
Notice: Scope(Class[Example_given_or_derived]): Ex1.maybe_half: 22
Notice: Scope(Class[Example_given_or_derived]): ex2.int: 44
Notice: Scope(Class[Example_given_or_derived]): ex2.maybe_half: 10.0
Notice: Scope(Class[Example_given_or_derived]): ex3.int: 12
Notice: Scope(Class[Example_given_or_derived]): ex3.maybe_half: 6
Notice: Scope(Class[Example_given_or_derived]): ex4.int: 12
Notice: Scope(Class[Example_given_or_derived]): ex4.maybe_half: 24.3
Notice: Compiled catalog for xxxx in environment production in 0.02 seconds
Notice: Applied catalog in 0.00 seconds
```

Note that I just pushed an update that simplifies the fix and ensures it works with both hash and positional argument initialization (I had only considered hash initialization in the first version).

I also tinkered with a patch that would have PCore enforce that a given value was returned in the event the Object attribute's Ruby method didn't, but while it was a good further learning experience with the code it wasn't quite as elegant as I would have liked and, more importantly, I'm not sure it is necessary.  I couldn't find anywhere in the spec where the PCore implementation is required to make sure the given value is returned so I think it is okay to trust the method author to return the value given (if given).

hlindberg:
Thought a bit more, I am not sure that "given or derived" is stated to guarantee that the given value is returned - imagine an attribute where you want it to be derived from the given value. No idea if that is very useful or if it needs to guarantee the return of a given.

seanmil:
> Thought a bit more, I am not sure that "given or derived" is stated to guarantee that the given value is returned - imagine an attribute where you want it to be derived from the given value. No idea if that is very useful or if it needs to guarantee the return of a given.

That's an excellent point - okay, I'll shelve the other branch but it sounds like this fix is good to go. Thanks so much for your review & feedback! Cheers!

hlindberg:
Actually- after more thought, the behavior of changing a given value is like having a "setter" method for an attribute (e.g. making a string all upper case, trimming white space, etc). The "given or derived" is more of a "getter" style method.
I therefore think it is a good idea to check that a given value is indeed returned and not some other getter/setter style computed value as that would be confusing.

PR #39, opened 2025-04-22
Update util.rb to fix closing of file.

adamboutcher:
When attempting to close files, the /proc/self loop was trying to close a file being converted to an integer which was causing issues on Fedora 42, this resolved the problem.

Also see https://github.com/puppetlabs/puppet/pull/9552

### DIFF
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -481,7 +481,7 @@ module Util
         Dir.foreach('/proc/self/fd') do |f|
           if f != '.' && f != '..' && f.to_i >= 3
             begin
-              IO.new(f.to_i).close
+              IO.new(f).close
             rescue
               nil
             end
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -479,9 +479,9 @@ module Util
 
       begin
         Dir.foreach('/proc/self/fd') do |f|
-          if f != '.' && f != '..' && f.to_i >= 3
+          if %{^\d+$}.match?(f) && f.to_i >= 3
             begin
-              IO.new(f).close
+              IO.new(f.to_i).close
             rescue
               nil
             end


adamboutcher:
For people searching this is the error that was flagging:

`/usr/share/ruby/vendor_ruby/puppet/util.rb:481:in 'Dir.foreach': Bad file descriptor - closedir (Errno::EBADF)`

silug:
If I understand this code correctly, it's trying to close each numbered file descriptor.  The `.to_i` is likely appropriate, but the file list should be filtered to only include names that consist entirely of digits.

silug:
I *think* what it is doing is closing all file descriptors other than stdin, stdout, stderr, and whatever fd 3 is.  :slightly_smiling_face: 

It does not use the full path in `/proc/self/fd`, just the numeric file descriptors.  On Linux, scanning `/proc/self/fd` is a handy way to do that.

adamboutcher:
> I _think_ what it is doing is closing all file descriptors other than stdin, stdout, stderr, and whatever fd 3 is. 🙂

I think that's what is intended but the to_i is possibly getting in the way...

I have Fedora 40 which does work (or silently fails) where as Fedora 42 gives the aforementioned error.
The modification in the PR works on both systems fine or is silently failing.

silug:
@adamboutcher Could you try adding `${^\d+$}.match?(f)` to the condition on line 482?  I suspect that will resolve the underlying problem.

The line should probably look like this:
```ruby
          if ${^\d+$}.match?(f) && f.to_i >= 3
```

silug:
To be clear, I think the change should look like this:
```diff
diff --git a/lib/puppet/util.rb b/lib/puppet/util.rb
index 06c7b11815..66be0f7da3 100644
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -479,7 +479,7 @@ module Util
 
       begin
         Dir.foreach('/proc/self/fd') do |f|
-          if f != '.' && f != '..' && f.to_i >= 3
+          if %{^\d+$}.match?(f) && f.to_i >= 3
             begin
               IO.new(f.to_i).close
             rescue
```
But I'm not seeing this problem for whatever reason, so I can't easily test it.

adamboutcher:
> ```diff
> if %{^\d+$}.match?(f) && f.to_i >= 3
> ```

I've tried this on Fedora42 and it seems to work, or at least I dont get any errors...
I'll change up and repush.

silug (approved):

nmburgan:
@ekohl Do you see any benefit to adding some of your logic from your other PR here? I think they both accomplish the goal. This one guards against *any* non-integer FD, but yours specifically filters out the ones we expect. I'm not sure if there are other edge cases we need to consider.

ekohl:
I don't think this actually resolves it. The file descriptor is the last number. If this avoids the error I think it won't close anything at all anymore. I'd suggest reverting this and applying my patch. See the linked bugzilla for minimal reproducer code

ekohl:
Based on https://bugzilla.redhat.com/show_bug.cgi?id=2349352#c9 a reproducer to show this broke the code:
```ruby
#!/usr/bin/env ruby

d = Dir.new('/proc/self/fd')
d.each_child do |f|
  if %{^\d+$}.match?(f) && f.to_i >= 3
    begin
      puts "Closing #{f}"
      IO.new(f.to_i).close
    rescue
      nil
    end
  else
    puts "Not closing #{f}"
  end
end
```

Running it shows it doesn't close anything anymore:

```console
$ ruby close1.rb
Not closing 0
Not closing 1
Not closing 2
Not closing 3
Not closing 4
Not closing 5
```

lib/puppet/util.rb
```diff
@@ -479,7 +479,7 @@ def safe_posix_fork(stdin = $stdin, stdout = $stdout, stderr = $stderr, &block)
 
       begin
         Dir.foreach('/proc/self/fd') do |f|
-          if f != '.' && f != '..' && f.to_i >= 3
+          if %{^\d+$}.match?(f) && f.to_i >= 3
```
ekohl, line 482: If you want this to work, it needs to be
```suggestion
          if %r{^\d+$}.match?(f) && f.to_i >= 3
```

Edit: but then you run into the original issue again, which is that the last file descriptor can't be closed.

ekohl:
I've opened https://github.com/OpenVoxProject/puppet/pull/42 to properly fix the issue.

Issue #41, opened 2025-04-25
[Feature request]: Fix file descriptor management in safe_posix_fork


jstraw:
### Use Case

From #39, we fixed a bug to only close integer file descriptors in /proc/self, but we should probably have a safer/standard way to track and close open fds that aren't standard in/out/error.

### Describe the solution you would like

We figure out what fd are open and close them without regex.

PR #42, opened 2025-05-01
Avoid closing directory we're iterating

ekohl:
This starts with reverting #39 because [it's broken](https://github.com/OpenVoxProject/puppet/pull/39#issuecomment-2840341655) and caused a regression. Then   I've taken my commits from https://github.com/puppetlabs/puppet/pull/9546.

Ruby 3.4 started error checking directory access and starts to raise Errno::EBADF.

This particular loop iterates on all open file descriptors and one is the directory listing from Dir.foreach.

In the past this could have led to leaked file descriptors, but it's unlikely since it's likely the last opened file descriptor and have the highest number.

Link: https://github.com/ruby/ruby/commit/f2919bd11c570fc5f5440d1f101be38f61e3d16b
Link: https://bugzilla.redhat.com/show_bug.cgi?id=2349352

### DIFF
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -479,7 +479,7 @@ module Util
 
       begin
         Dir.foreach('/proc/self/fd') do |f|
-          if %{^\d+$}.match?(f) && f.to_i >= 3
+          if f != '.' && f != '..' && f.to_i >= 3
             begin
               IO.new(f.to_i).close
             rescue
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -481,12 +481,12 @@ module Util
         d = Dir.new('/proc/self/fd')
         ignore_fds = ['.', '..', d.fileno.to_s]
         d.each_child do |f|
-          if !ignore_fds.include?(f) && f.to_i >= 3
-            begin
-              IO.new(f.to_i).close
-            rescue
-              nil
-            end
+          next if ignore_fds.include?(f) || f.to_i < 3
+
+          begin
+            IO.new(f.to_i).close
+          rescue
+            nil
           end
         end
       rescue Errno::ENOENT, Errno::ENOTDIR # /proc/self/fd not found, /proc/self not a dir
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -478,8 +478,10 @@ module Util
       $stderr = STDERR
 
       begin
-        Dir.foreach('/proc/self/fd') do |f|
-          if f != '.' && f != '..' && f.to_i >= 3
+        d = Dir.new('/proc/self/fd')
+        ignore_fds = ['.', '..', d.fileno.to_s]
+        d.each_child do |f|
+          if !ignore_fds.include?(f) && f.to_i >= 3
             begin
               IO.new(f.to_i).close
             rescue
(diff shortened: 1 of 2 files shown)


lib/puppet/util.rb
```diff
@@ -478,13 +478,15 @@ def safe_posix_fork(stdin = $stdin, stdout = $stdout, stderr = $stderr, &block)
...
-          if %{^\d+$}.match?(f) && f.to_i >= 3
-            begin
-              IO.new(f.to_i).close
-            rescue
-              nil
-            end
+        d = Dir.new('/proc/self/fd')
+        ignore_fds = ['.', '..', d.fileno.to_s]
+        d.each_child do |f|
+          next if ignore_fds.include?(f) || f.to_i < 3
```
ekohl, line 484: Looking at #41 I wonder if we should explicitly exclude the FDs that we want to keep open:
```suggestion
        ignore_fds = ['.', '..', stdin.fileno.to_s, stdout.fileno.to_s, stderr.fileno.to_s, d.fileno.to_s]
        d.each_child do |f|
          next if ignore_fds.include?(f)
```

This version is probably safer. Not sure if it's valid to have no stdin, stdout or stderr.
```suggestion
        ignore_fds = ['.', '..'] + [stdin, stdout, stderr, d].filter_map { |f| f&.fileno&.to_s }
        d.each_child do |f|
          next if ignore_fds.include?(f)
```

What makes me hesitant is that I don't know

ekohl:
And I just noticed this repo doesn't run any of the tests, but when I locally run it I did break them:
https://github.com/OpenVoxProject/puppet/blob/13dd914bfd188af26a4acabe234a5b292188fd88/spec/unit/util_spec.rb#L597-L665

lib/puppet/util.rb
nmburgan: I'm a little worried we're playing whack-a-mole with things we want to filter out and I'm not sure how to be sure we're being exhaustive about it.

ekohl: IMHO this is much better than looking at `.to_i < 3` and assuming that those file descriptors are stdin, stdout and stderr since it is explicit about what's going on.

nmburgan: Agree it's better than `< 3`. Just wonder if we have any other gaps of things we should be ignoring.

ekohl: I don't think so.

Ideally we would use `open3` instead of `fork()`. That can change the user, cwd, set env vars etc but that's a larger refactor I don't want to take on right now.

nmburgan:
@ekohl Have you been able to test this change and make sure it doesn't break anything? Looks okay to me, but I'm not super knowledgeable about this area.

ekohl:
It's already rolled out in Fedora 42 and I wrote a test there. Without the patch anything that execs (like the exec resource) just straight up fails. With the patch it works again. See the linked bugzilla for more details.  The micro-reproducers there give confidence it's correct

jstraw:
https://docs.fedoraproject.org/en-US/defensive-coding/tasks/Tasks-Descriptors/#sect-Defensive_Coding-Tasks-Descriptors-Child_Processes talks about methodologies for this exact thing.  It looks like ruby for some reason doesn't use FD_CLOEXEC/O_CLOEXEC to ensure they close on fork end.

ekohl:
The Fedora page mentions that has a race condition. This implements the approach 2 it mentions: loop over `/proc/self/fd` with a fallback to static iteration.

My suggestion of using open3 would use a better system call to avoid all thr code, but is a much bigger refactor

nmburgan (approved):

PR #60, opened 2025-05-13
CI: Use builtin option to install gems

bastelfreak:


bastelfreak:
why does the rubocop job fail with `Unable to find gem voxpupuli-rubocop; is the gem installed? Gem::MissingSpecError` :thinking:

.github/workflows/checks.yaml
```diff
@@ -16,26 +16,19 @@ jobs:
...
 
       - name: Install ruby version ${{ matrix.cfg.ruby }}
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.cfg.ruby }}
-
-      - name: Update rubygems and install gems
-        run: |
-          gem update --system --silent --no-document
-          bundle config set without packaging documentation
```
ekohl, line 37: You can set `BUNDLE_WITHOUT=packaging:documentation` to skip those groups

PR #61, opened 2025-05-13
Update rubocop

nmburgan:
For the moment, this continues to shove linting issues under the rug. Will look at pulling them out from under the rug in another PR.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,13 +1,14 @@
 inherit_from: .rubocop_todo.yml
 
-require:
+plugins:
   - rubocop-i18n
   - rubocop-performance
   - rubocop-rake
   - rubocop-rspec
 
 AllCops:
-  TargetRubyVersion: 3.1
+  TargetRubyVersion: 3.2
+  NewCops: enable
   Include:
     - 'lib/**/*.rb'
     - 'ext/**/*.rb'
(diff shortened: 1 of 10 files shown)


.rubocop.yml
```diff
@@ -1,13 +1,14 @@
...
-require:
+plugins:
   - rubocop-i18n
   - rubocop-performance
   - rubocop-rake
   - rubocop-rspec
 
 AllCops:
-  TargetRubyVersion: 3.1
+  TargetRubyVersion: 3.2
```
bastelfreak: If we do this, we would need to bump `required_ruby_version` in the gemspec as well. I never understood why Perforce decided to set ruby 3.1 as lower version but vendor Ruby 3.2 in the AIO.

nmburgan: Hmm, yeah, I suppose we should.

nmburgan: Ah, I think it's because JRuby 9.4 only supports 3.1.  But 3.1 is EOL so that's why 3.2 is in the AIO.  JRuby 10 supports 3.4, but that's a lot more work to get there.  I should probably remove the `set` commit since that might not work with JRuby.

bastelfreak: Ah maybe that's why it's pinned to 3.1 :( IMO we should stick to that version until we do a major release.

PR #63, opened 2025-05-17
Add 'implementation' fact to agent and server.

ffrank:
As specified in OpenVoxProject/planning#39, the agent will now make its implementation flavor known to the compiler through a built-in fact, as will the server (through a server fact).

### DIFF
--- a/lib/puppet/node/facts.rb
+++ b/lib/puppet/node/facts.rb
@@ -28,6 +28,7 @@ class Puppet::Node::Facts
   attr_accessor :name, :values, :timestamp
 
   def add_local_facts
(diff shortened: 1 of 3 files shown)


ffrank:
The testing for the server_facts is..."limited".

Looks like there are only a bunch of tests in the unit[sic!] tests for the [compiler terminus](https://github.com/OpenVoxProject/puppet/blob/6775c8ceabc9ef1ca92395bf35d2492ad57ee61e/spec/unit/indirector/catalog/compiler_spec.rb#L461). (These look more like integration tests of the second degree.)

PR #66, opened 2025-05-20
Rename gemspec to match gem name

bastelfreak:


bastelfreak:
lets do https://github.com/OpenVoxProject/puppet/pull/67 instead

PR #67, opened 2025-05-21
Support both puppet and openvox gem names

ekohl:
This uses the filename to determine the gem name. This allows users to choose:

```ruby
gem 'puppet', github: 'OpenVoxProject/puppet'
```

Or use the new openvox gem:

```ruby
gem 'openvox', github: 'OpenVoxProject/puppet'
```

At some point they can start to diverge and make puppet more of a transitional gem that only depends on openvox.

Currently untested, but this allows for discussion.

### DIFF
--- a/openvox.gemspec
+++ b/openvox.gemspec
@@ -1,0 +1,1 @@
+puppet.gemspec
--- a/puppet.gemspec
+++ b/puppet.gemspec
@@ -1,5 +1,5 @@
 Gem::Specification.new do |spec|
-  spec.name = "openvox"
+  spec.name = File.basename(__FILE__, '.gemspec')
   spec.version = "8.18.1"
   spec.licenses = ['Apache-2.0']
 


openvox.gemspec
```diff
@@ -0,0 +1 @@
+puppet.gemspec
```
bastelfreak, line 1: I would prefer if if the openvox gemspec is the actual one, but that probably doesn't matter.

ekohl: I thought about that too with the eye on the future where `puppet.gemspec` may be an empty spec that only depends on `openvox`. We can still do that.

bastelfreak: Mhm true, we can always flip it.

bastelfreak (approved):

bastelfreak:
works fine:

```
# for gemspec in puppet.gemspec openvox.gemspec; do gem build --strict --verbose $gemspec; done
  Successfully built RubyGem
  Name: puppet
  Version: 8.18.1
  File: puppet-8.18.1.gem
  Successfully built RubyGem
  Name: openvox
  Version: 8.18.1
  File: openvox-8.18.1.gem
```

bastelfreak:
@ekohl should we merge this, and afterwards I update the release action to publish the puppet gem also to the github registry?

bastelfreak:
mhm

```
$ bundle install --path .vendor --with development:release:test

[!] There was an error parsing `Gemfile`: There are multiple gemspecs at /home/bastelfreak/code/openvox/puppet. Please use the :name option to specify which one should be used. Bundler cannot continue.

 #  from /home/bastelfreak/code/openvox/puppet/Gemfile:3
 #  -------------------------------------------
 #
 >  gemspec
 #
 #  -------------------------------------------
```

bastelfreak:
Does it maybe make more sense to move the second gemspec into another directory? Then we don't need to always specificy the gemspec we're using.

bastelfreak:
Ah it's not that complicated:

```patch
diff --git a/Gemfile b/Gemfile
index 5fa0ec958f..828a132282 100644
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"

-gemspec
+gemspec name: 'openvox'

 def location_for(place, fake_version = nil)
   if place.is_a?(String) && place =~ /^((?:git[:@]|https:)[^#]*)#(.*)/
```

This tells bundler/Gemfile to use the openvox.gemspec

With that, bundle works:

```
$ bundle install --path .vendor --with development:release:test
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Using openvox 8.18.1 (was 8.17.0) from source at `.` and installing its executables
Bundle complete! 39 Gemfile dependencies, 110 gems now installed.
Bundled gems are installed into `./.vendor`
```

bastelfreak:
> Publishing shouldn't be needed either.

yes we don't have to. But I think it would be nice for users from a convenience point of view.

ekohl:
After further thinking and testing I think it's easier if the Vox ecosystem switches completely. Some open PRs to do so:

I think there are a few gems:
* [puppet-syntax](https://github.com/voxpupuli/puppet-syntax): switch to depend on puppet https://github.com/voxpupuli/puppet-syntax/pull/187
* [puppet-strings](https://github.com/puppetlabs/puppet-strings): forked to [openvox-strings](https://github.com/voxpupuli/openvox-strings) with https://github.com/voxpupuli/openvox-strings/pull/2
* [voxpupuli-release](https://github.com/voxpupuli/voxpupuli-release) needs to use openvox-strings: https://github.com/voxpupuli/voxpupuli-release/blob/e8f760b74e99f4c1eefdfde9b920893903c24f52/voxpupuli-release.gemspec#L24-L28
* [voxpupuli-test](https://github.com/voxpupuli/voxpupuli-test) needs to use openvox-strings: https://github.com/voxpupuli/voxpupuli-test/blob/49e0e8a6a9cff4708034b3cb55b045685ee2bfa6/voxpupuli-test.gemspec#L24

Then there is [rspec-puppet](https://github.com/puppetlabs/rspec-puppet) which needs to be switched over. Probably also by forking to `rspec-openvox`.

Once we've done that we also need to audit the server because some tooling there also pulls in the puppet gem.

PR #84, opened 2025-05-30
Release 8.19.0

smortex:


bastelfreak:
We updated some gem constraints to allow newer version, like https://github.com/OpenVoxProject/puppet/pull/77. I'm a bit unsure if we should include those or not in the changelog.

bastelfreak:
speaking about it, I also just merged https://github.com/OpenVoxProject/puppet/pull/52

smortex:
Should these affect end-users?  I trend to trim what is visible in the changelog to what affect end users and remove anything that is "internal" or "implementation detail".

It looks like the configuration we have for CHANGELOG.md and the release is not 1-1 matching, as PRs labelled "dependencies" will now appear in the latter after #83.

smortex:
> That's why I'm unsure :D I makes a difference for all the more techy users, like us, because the whole gem dependencies are now more relaxed and it's now possible to install other tools from the ecosystem. But I don't mind it if we keep it out of the changelog, I just wanted to mention it.

If we change our mind, we can relabel PRs and on next release the ChangeLog will be updated with the new content, so I think is is not a major issue and we may have some feedback to decide what to do exactly?



> Also I'm not sure if the whole vanagon tooling will create the git tag in the puppet repo, or if we should/could create it 🤔

My understanding is that you release a new tag, and point to it in the right place (the commit since a tag can be changed, e.g. https://github.com/OpenVoxProject/openvox-agent/blob/a28a93c53ca617b0a0ce75a5a3ae91d1f5c1bf50/configs/components/puppet.json#L1) before moving on with building the package.

bastelfreak (approved):

PR #87, opened 2025-05-31
Add implementation fact to agent and server

bastelfreak:


bastelfreak:
backported from https://github.com/OpenVoxProject/puppet/pull/78

PR #88, opened 2025-05-31
Generate human-readable ChangeLog

bastelfreak:
backported from https://github.com/OpenVoxProject/puppet/pull/82


Rakefile
```diff
@@ -131,3 +131,26 @@ if Rake.application.top_level_tasks.grep(/^gettext:/).any?
...
+    config.header = <<~HEADER.chomp
+    # Changelog
+
+    All notable changes to this project will be documented in this file.
+    HEADER
+    config.user = "openvoxproject"
+    config.project = "puppet"
+    config.exclude_labels = %w[dependencies duplicate question invalid wontfix wont-fix modulesync skip-changelog]
+    config.future_release = Puppet::PUPPETVERSION
+    config.since_tag = '7.36.1'
```
smortex, line 149: If we `require_relative 'lib/puppet/version'`, we can:

```suggestion
    config.since_tag = Puppet::PUPPETVERSION
```

CHANGELOG.md
```diff
@@ -0,0 +1,52 @@
...
+- Bad link for Github Issues [\#26](https://github.com/OpenVoxProject/puppet/issues/26)
+
+**Merged pull requests:**
+
+- dependabot: check for github actions and bundler [\#46](https://github.com/OpenVoxProject/puppet/pull/46) ([bastelfreak](https://github.com/bastelfreak))
+- Reenable and cleanup CI configuration [\#45](https://github.com/OpenVoxProject/puppet/pull/45) ([bastelfreak](https://github.com/bastelfreak))
+- Remove unused ContentsDescription class [\#44](https://github.com/OpenVoxProject/puppet/pull/44) ([ekohl](https://github.com/ekohl))
+- Fix a typo and some formatting in the type and provider docs [\#43](https://github.com/OpenVoxProject/puppet/pull/43) ([smoeding](https://github.com/smoeding))
+- Avoid closing directory we're iterating [\#42](https://github.com/OpenVoxProject/puppet/pull/42) ([ekohl](https://github.com/ekohl))
+- Set version to 8.16.0 [\#40](https://github.com/OpenVoxProject/puppet/pull/40) ([nmburgan](https://github.com/nmburgan))
```
smortex, line 31: This should not appear in the 7.x changelog ;-)

puppet.gemspec
```diff
@@ -1,6 +1,8 @@
+require_relative 'lib/puppet/version'
```
smortex, line 1: This happened to be a terrible idea as the file is copied in AIO packages while it is used to generate the final .gemspec file when installing from a gem and the path do not match.

PR #95, opened 2025-06-03
Borrow the release workflow from the main branch

smortex:
While here, adjust the GitHub registry URL to test the chages in #86.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,41 @@
+---
+name: Gem Release
+
+on:
+  push:
+    tags:
+      - '7.*'
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
+---
+name: Gem Release
+
+on:
+  push:
+    tags:
+      - '*'
```
bastelfreak: Should we add the filter, just in case?

```suggestion
      - 7.**
```

smortex: :+1:

bastelfreak (approved):

PR #98, opened 2025-06-04
Add base64 and syslog as gem dependencies for Ruby 3.4

ekohl:
With Ruby 3.4 the base64 gem changed from a default gem to a bundled gem. This means it needs to be specified as a dependency.


bastelfreak (approved):

bastelfreak (approved):

bastelfreak:
reverted syslog in https://github.com/OpenVoxProject/puppet/pull/110, so adjusted the title

PR #102, opened 2025-06-05
GCG: move it to dedicated gem group

bastelfreak:
That allows us to skip the installation. It also has dependencies that are incompatible with JRuby.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,7 +68,6 @@ group(:development, optional: true) do
 end
 
 group(:packaging) do
-  gem 'github_changelog_generator'
   gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
 end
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -68,7 +68,6 @@ group(:development, optional: true) do
 end
 
 group(:packaging) do
-  gem 'github_changelog_generator'
   gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
```
ekohl, line 71: This also has many dependencies. Can we make the whole packaging group optional too?

Overall I think we should carefully review the gem because it comes from Perforce. It's likely we don't need most of it.

ekohl (changes requested):
Overall :+1: for not installing it by default.

In `Rakefile` when the gem can't be loaded it mentions that you need to install with the packaging group. That would no longer be true after this.

Gemfile
bastelfreak: I don't know enough about it, so I didn't want to touch it yet.

bastelfreak:
@ekohl fixed the Rakefile

ekohl (approved):

Gemfile
ekohl: It only provides packaging tasks and is lazily loaded. I think it's safe to mark it optional as well

bastelfreak: I will test a bit and raise another PR for it

PR #106, opened 2025-06-05
Fix CI on JRuby

smortex:


ekohl:
We ended up dropping the dependency in https://github.com/OpenVoxProject/puppet/pull/110

PR #116, opened 2025-06-06
Release: Fix path to gem

bastelfreak:
The gem artifact isn't in the pkg/ subfolder, it's in the root dir.


bastelfreak:
need to be backported to 7.x as well

jay7x (approved):

smortex (approved):
Ah! lol

PR #120, opened 2025-06-13
Gh 101 openvox acceptance gha main

jpartlow:


jpartlow:
Waiting for https://github.com/jpartlow/puppet/actions/runs/15642839849/job/44074051996 to finish grinding through.

Ignore the Debian 13 failure. We don't have openvox-server debian 13 packages yet, and I've commented out that plaform in 5cf7c0ee5.

jpartlow:
Ok, there are definitely some test failures I need to look into from that Beaker run. Once that run completes I'll grab the log set and sift through them.

jpartlow:
Ok, have a passing run now with the additional changes: https://github.com/jpartlow/puppet/actions/runs/15741979892/job/44369321462

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,208 @@
...
+          - [rocky, '8']
+          - [rocky, '9']
+          - [ubuntu, '18.04']
+          - [ubuntu, '20.04']
+          - [ubuntu, '22.04']
+          - [ubuntu, '24.04']
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
bastelfreak: Did you set that on purpose? do we need the full history?

jpartlow: We shouldn't...I'm not sure exactly where that crept in, actually, so probably a copy/paste. I'll remove it.

jpartlow: Removed!

bastelfreak (approved):

PR #121, opened 2025-06-13
Fix Tests for JRuby 9.4

bastelfreak:


spec/unit/parser/compiler_spec.rb
```diff
@@ -874,7 +874,10 @@ class #{name} {}
               include #{name}
             }
           MANIFEST
-          }.to raise_error(Puppet::Error, /Class '#{name}' is already defined \(line: 1\); cannot be redefined as a node \(line: 2\) on node #{name}/)
+          # error message differs a bit
+          # jruby 9.4: Class 'someone' is already defined (line: 1); cannot be redefined as a node (line: 2) (line: 2) on node someone
```
bastelfreak, line 878: I'm not sure if this is an actual bug. It seems odd to output the line number twice, but I've no idea where this is coming from.

PR #138, opened 2025-06-19
Add prepare_release / release workflows

smortex:
These github workflows can be run manually by any contributor with write access to the repository to prepare and create relases of this component, making the release process less prone to mistakes.

This rely on the tasks that where added to the shared-actions repository:
* https://github.com/OpenVoxProject/shared-actions/pull/11
* https://github.com/OpenVoxProject/shared-actions/pull/12


.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+        description: 'The branch where we do this release.'
+        required: false
+        default: ''
+        type: string
+
+permissions: {}
+
+jobs:
+  release:
+    uses: OpenVoxProject/shared-actions/.github/workflows/release.yml@main
```
smortex, line 21: This has not been merged into main yet.  Will be functional when https://github.com/OpenVoxProject/shared-actions/pull/12 is merged.

tasks/vox.rake
smortex, line 1: This has the content of `tasks/tag.rake` (file which is removed), adjusted to only to the version bump.  Unfortunately GitHub does not detect it as a renamed file that was modified, and do not render a convenient diff.

.github/workflows/release.yml
bastelfreak: merged https://github.com/OpenVoxProject/shared-actions/pull/12

bastelfreak (approved):

PR #142, opened 2025-06-27
Switch from facter to openfact

smortex:
Now that there is an openfact gem, switch to it and let's eat our own dog food!


Gemfile
```diff
@@ -14,9 +14,9 @@ end
 
 # Make sure these gem requirements are in sync with the gempspec. Specifically,
 # the runtime_dependencies in openvox.gemspec match the runtime dependencies here
-# (like facter, semantic_puppet, and puppet-resource_api)
+# (like openfact, semantic_puppet, and puppet-resource_api)
 
-gem "facter", *location_for(ENV['FACTER_LOCATION'] || ["~> 4.3"])
+gem "openfact", *location_for(ENV['OPENFACT_LOCATION'] || [">= 5.0.0.rc0", "< 6"])
```
bastelfreak: I think before we merge this we should make a 5.0.0 release.

smortex: Agreed, we will also remove the `.rc0`.

I wanted to be sure it does not explode for some reason when switching, which seems to be fine :tada:.

bastelfreak (approved):

PR #147, opened 2025-07-09
Remove facter from install.rb

ekohl:
This drops the use of Facter in `install.rb`. It drops Solaris support and then rewrites the Windows detect to use `RUBY_PLATFORM` instead. It then also drops the prereq checking since that doesn't really serve a purpose anymore.

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -38,8 +38,6 @@ require 'ostruct'
 
 IS_WINDOWS = RUBY_PLATFORM.match?(/(mswin|mingw)/)
 
-PREREQS = %w{openssl cgi}
-
 InstallOptions = OpenStruct.new
 
 def glob(list)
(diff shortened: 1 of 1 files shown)


install.rb
```diff
@@ -36,8 +36,9 @@
 require 'optparse'
 require 'ostruct'
 
-PREREQS = %w{openssl facter cgi}
-MIN_FACTER_VERSION = 1.5
+IS_WINDOWS = RUBY_PLATFORM.match?(/(mswin|mingw)/)
```
smortex, line 39: Will not rubocop if we enable it one day complain that we should freeze constants?  It is probably something we will do today, so probably not a big deal for now.

smortex (approved):

install.rb
ekohl: There's a lot more to rubocop in this file ;)

PR #151, opened 2025-07-10
Consistently use `.yml` extension for YAML files

smortex:


.github/workflows/gem_release.yml
smortex, line 1: Needs to be reflected on rubygem.org

PR #154, opened 2025-07-22
(ci) Use shared_actions/.github/workflows/beaker_acceptance.yml

jpartlow:
Replaces the acceptance.yaml's jobs with a call to the reusable
beaker_acceptance.yml workflow from the openvoxproject/shared_actions
repository.

This version also takes a ref input and should be able to test 7.x
suite/packages as well without switching to the 7.x branch.

---

Also renames acceptance.yaml to acceptance.yml for consistency with other workflows/projects.


.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,120 @@
...
+          true)
+        default: 'https://s3.osuosl.org/openvox-artifacts'
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  acceptance:
+    uses: 'jpartlow/shared-actions/.github/workflows/beaker_acceptance.yml@add-reusable-beaker-acceptance-workflow'
```
jpartlow: This needs to be updated after https://github.com/OpenVoxProject/shared-actions/pull/18 is merged.

jpartlow: Fixed.

bastelfreak (approved):

PR #156, opened 2025-07-23
Make passing invalid options to install.rb fatal

ekohl:
In b909b673bae784af7822fcd58549fd6773b4320b the --[no-]prereq-checks option was dropped. The error was printed, but then continued to execute and it ignored all passed options. That caused a regression in 8.21.0.

It now catches the error and exits with a non-zero exit code, forcing the packager to deal with it.

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -159,6 +159,9 @@ def prepare_installation
     end
 
     opts.parse!
+  rescue OptionParser::InvalidOption => e
+    $stderr.puts e
+    exit 1
   end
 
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

ekohl:
Just noticed a typo in the release: it was 8.21.0, not 8.20.0. No content change, just fixed up the commit message.

ekohl:
For Windows failures I've opened https://github.com/OpenVoxProject/openvox/pull/159.

PR #157, opened 2025-07-23
CI: add dummy job we can depend on as required check

bastelfreak:


ekohl (approved):
There's probably more we could clean up, but this at least allows us to make it a required check and enable auto merging.

.github/workflows/tests.yaml
```diff
@@ -85,3 +108,12 @@ jobs:
...
           bundle exec rake parallel:spec
+
+  tests:
+    needs:
+      - checks
+      - rspec_tests
+    runs-on: ubuntu-24.04
+    name: Test suite
+    steps:
+      - run: echo Test suite completed
```
ekohl: On second thought: this should use https://github.com/re-actors/alls-green because with skipped jobs it can give false positives. See https://github.com/theforeman/actions/blob/2a5671906614405a7480f53403d09f6db33816bb/.github/workflows/test-gem.yml#L88-L98 for an example.

ekohl (changes requested):
Sorry for flip-flopping :(

.github/workflows/tests.yaml
bastelfreak: Is that really required here? All of our checks are defined in this file and they should never be skipped. And the backport job is triggered manually (and we should probably remove it, we inherited it from Perforce and it isn't working).

bastelfreak: https://github.com/OpenVoxProject/openvox/pull/158

ekohl: > All of our checks are defined in this file and they should never be skipped. And the backport job is triggered manually (and we should probably remove it, we inherited it from Perforce and it isn't working).

Pure `needs` on a check is insufficient and this needs to always run. If a `needs` fails this may end up being skipped and then the required check isn't present. You really want to use this action and we do the same in gha-puppet: https://github.com/voxpupuli/gha-puppet/blob/b3636f92bc554e97a00b2513546db16c5035be4b/.github/workflows/basic.yml#L109-L118 & https://github.com/voxpupuli/gha-puppet/blob/b3636f92bc554e97a00b2513546db16c5035be4b/.github/workflows/beaker.yml#L207-L218

bastelfreak: done

ekohl (approved):

PR #161, opened 2025-07-26
Fix jruby tests

sebastianrakel:


spec/unit/http/service/compiler_spec.rb
```diff
@@ -594,7 +594,7 @@
       invalid_facts = Puppet::Node::Facts.new(certname, {'invalid_utf8_sequence' => "\xE2\x82".force_encoding('binary')})
       expect {
         subject.put_facts(certname, environment: 'production', facts: invalid_facts)
-      }.to raise_error(Puppet::HTTP::SerializationError, /Failed to serialize Puppet::Node::Facts to json: ("\\xE2" from ASCII-8BIT to UTF-8|partial character in source, but hit end)/)
```
bastelfreak, line 597: I don't know yet what would have raised `"\\xE2" from ASCII-8BIT to UTF-8`. `partial character in source, but hit end` should come from the json gem java version: https://github.com/ruby/json/blob/master/java/src/json/ext/StringDecoder.java#L168-L176.

Maybe that exception just isn't raised anymore :thinking:

spec/unit/util/json_spec.rb
```diff
@@ -23,7 +23,7 @@
     it 'raises an error if JSON is invalid' do
       expect {
         Puppet::Util::Json.load('{ invalid')
-      }.to raise_error(Puppet::Util::Json::ParseError, /expected object key, got 'invalid' at line 1 column 3/)
```
bastelfreak, line 26: This also looks like an error from the json lib:

```
bastelfreak@bastelfreak-nb ~/code/json $ git grep 'expected object key'
ext/json/ext/parser/parser.c:                    raise_parse_error("expected object key, got %s", state);
ext/json/ext/parser/parser.c:                            raise_parse_error("expected object key, got: %s", state);
test/json/json_ext_parser_test.rb:      assert_equal "expected object key, got EOF at line 1 column 5", ex.message
bastelfreak@bastelfreak-nb ~/code/json $
```

bastelfreak: Hahaha. While digging around I found https://github.com/ruby/json/issues/754, which I raised some months ago. From the developers:

> So it's probably best if your test only assume a specific error class, but doesn't try to match the message content.

Still it would be nice to figure out it this passes with an older json gem version.

jay7x:
Had a look at jruby 9.4.6.0 changelog and they did some changes around exceptions handling and printing (https://github.com/jruby/jruby/pull/8040 e.g.).
Also I did quick googling and found no reports about duplicate file & line position info in the exceptions.. so I guess it's something in the code we own now

bastelfreak:
@jay7x given this is also the current state in PE, I think it's fine to merge it.

bastelfreak (approved):

PR #170, opened 2025-08-21
Allow file content that looks like a checksum.

### LINKED ISSUE #169: [Bug]: lines in file resource content looking like a checksum throws an error

### Describe the Bug

If I try to create a file with content that looks like this : `puppet apply -e "file { '/tmp/test-bug-puppet': ensure => 'file', content => '{ABCD}X' }"`

a warning and an error are displayed and the file is not created.

### Expected Behavior

A file is created with content `{ABCD}X`

### Steps to Reproduce

`puppet apply -e "file { '/tmp/test-bug-puppet': ensure => 'file', content => '{ABCD}X' }"`

### Environment

Version: 8.19.2
Platform: Debian 12

### Additional Context

The same problem was filed against Puppet 8.7 : https://github.com/puppetlabs/puppet/issues/9412

### Relevant log output

```shell
Warning: Using a checksum in a file's "content" property is deprecated. The ability to use a checksum to retrieve content from the filebucket using the "content" property will be removed in a future release. The literal value of the "content" property will be written to the file. The checksum retrieval functionality is being replaced by the use of static catalogs. See https://puppet.com/docs/puppet/latest/static_catalogs.html for more information.
   (file: unknown, line: 1)
Error: Could not retrieve content for {ABCD}X from filebucket: Could not get filebucket from file
Error: /Stage[main]/Main/File[/tmp/test-bug-puppet]/ensure: change from 'absent' to 'file' failed: Could not retrieve content for {ABCD}X from filebucket: Could not get filebucket from file
```

jeremie-pierson:
_This pull request is intended to fix #169 . Also this is my first pull request on github._

Currently, if a file has a content attribute set to something that looks like a checksum, it will display a deprecation warning and then probably throw an error, as the checksum-like string won't match anything in filebuckets.

This code is apparently intended to allow a checksum to be passed instead of actual content, with the effect of replacing file content if it doesn't match the checksum. It appears that this mecanism is replaced by "static catalogs" and was scheduled for removal in Puppet 7.

As it is no longer documented (because deprecated), it is surprising to stumble upon the behavior by just having file content that looks like a checksum. I had to work around this in a real usecase involving some proprietary software that uses the same syntax for values to be encrypted at startup.

This commit introduces a new setting "use_checksum_in_file_content" that default to true, preserving current behavior. If set to false, it will never look for checksums in file contents.

This setting should probably be set to false by default in the next major release.

### DIFF
--- a/lib/puppet/type/file/content.rb
+++ b/lib/puppet/type/file/content.rb
@@ -54,27 +54,6 @@ module Puppet
         if @actual_content || resource.parameter(:source)
           # Actual content is already set, value contains it's checksum
           value
-        elsif Puppet[:use_checksum_in_file_content]
-          # The value passed in the "content" attribute of this file looks like
-          # a checksum, and this is intended by the user.
-          # Display a warning as this behavior is deprecated.
-          Puppet.puppet_deprecation_warning([
-            # TRANSLATORS "content" is an attribute and should not be translated
-            _('Using a checksum in a file\'s "content" property is deprecated.'),
-            # TRANSLATORS "filebucket" is a resource type and should not be translated. The quoted occurrence of "content" is an attribute and should not be translated.
-            _('The ability to use a checksum to retrieve content from the filebucket using the "content" property will be removed in a future release.'),
-            # TRANSLATORS "content" is an attribute and should not be translated.
-            _('The literal value of the "content" property will be written to the file.'),
-            # TRANSLATORS "static catalogs" should not be translated.
-            _('The checksum retrieval functionality is being replaced by the use of static catalogs.'),
-            _('See https://puppet.com/docs/puppet/latest/static_catalogs.html for more information.')
-          ].join(" "),
-                                            :file => @resource.file,
-                                            :line => @resource.line) if !@actual_content && !resource.parameter(:source)
-          # We return the value assuming it really is the checksum of the
-          # actual content we want. It should be fetched from filebucket
-          # later on.
-          value
         else
           # The content only happens to look like a checksum by chance.
           @actual_content = value.is_a?(Puppet::Pops::Types::PBinaryType::Binary) ? value.binary_buffer : value
@@ -172,17 +151,11 @@ module Puppet
     def each_chunk_from
       if actual_content.is_a?(String)
         yield actual_content
-      elsif content_is_really_a_checksum? && actual_content.nil?
+      elsif actual_content.nil?
         yield read_file_from_filebucket
       elsif actual_content.nil?
         yield ''
       end
-    end
-
-    def content_is_really_a_checksum?
-      return false unless Puppet[:use_checksum_in_file_content]
-
-      checksum?(should)
     end
 
     def read_file_from_filebucket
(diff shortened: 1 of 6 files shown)


jeremie-pierson:
I'll correct the new rubocop offense.

binford2k:
in all honesty, this was deprecated seven years ago and I've never seen anyone actually use it. I vote we just remove the functionality.

jeremie-pierson:
> in all honesty, this was deprecated seven years ago and I've never seen anyone actually use it. I vote we just remove the functionality.

OK, I'll update this PR to remove the functionality instead of adding another config item :-)

lib/puppet/type/file/content.rb
```diff
@@ -48,23 +48,19 @@ module Puppet
...
+        # attribute in Puppet code, that happens to look like a checksum, or is
+        # it an actual checksum computed on the actual content?
+        if @actual_content || resource.parameter(:source)
+          # Actual content is already set, value contains it's checksum
+          value
+        else
+          # The content only happens to look like a checksum by chance.
+          @actual_content = value.is_a?(Puppet::Pops::Types::PBinaryType::Binary) ? value.binary_buffer : value
+          resource.parameter(:checksum).sum(@actual_content)
+        end
```
silug: I think this can be removed.

lib/puppet/type/file/content.rb
```diff
@@ -155,17 +151,13 @@ def write(file)
     def each_chunk_from
       if actual_content.is_a?(String)
         yield actual_content
-      elsif content_is_really_a_checksum? && actual_content.nil?
+      elsif actual_content.nil?
         yield read_file_from_filebucket
```
silug: This can be removed.

silug (changes requested):

lib/puppet/type/file/content.rb
jeremie-pierson: I tried without it but got obscure errors. I think the problem was that the checksum mechanism is not only used for the functionality we're trying to remove, but also for other purposes in the file handling code. Probably has to do with filebuckets.

It seems like "value" can legitimately contain a checksum if this attribute ("content") was already set. I'm not sure about this, but as I understand it, since we are in a "munge" block, we're transforming the value of the attribute prior to storing it. In the standard case, when user sets the "content" attribute to something in Puppet code, we store the actual content in `@actual_content` and attribute value holds the checksum.

As I recall, when I first removed theses lines, I had loads of errors in filebucket-related code elsewhere because, presumably, other code assumes that file content attribute value is in fact a checksum...

If you think I'm mistaken, I'll try to delete those lines again and see what CI says this time.

lib/puppet/type/file/content.rb
jeremie-pierson: This could be related to above discussion (or I could very well be mistaken about all this :-) ).

bastelfreak:
@jeremie-pierson please rebase against our main branch to get rid of the merge commit. And please sign your commit with a gpg or ssh key.

silug (approved):

bastelfreak:
(removing the enhancement label, otherwise the changelog generator gets confused)

bastelfreak (approved):

PR #171, opened 2025-08-22
Ensure confdir exists

binford2k:
Prior to this, if OpenVox was installed via gem and the first command you run is setting configuration then it fails because `confdir` doesn't exist yet. This corrects that.

It doesn't bother setting permissions because future runs will correct the permissions.

### DIFF
--- a/lib/puppet/face/config.rb
+++ b/lib/puppet/face/config.rb
@@ -171,6 +171,7 @@ Puppet::Face.define(:config, '0.0.1') do
       end
 
       path = Puppet::FileSystem.pathname(Puppet.settings.which_configuration_file)
+      Puppet::FileSystem.dir_mkpath(path)
       Puppet::FileSystem.touch(path)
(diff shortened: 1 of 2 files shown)


lib/puppet/face/config.rb
```diff
@@ -171,6 +171,7 @@ def report_section_and_environment(section_name, environment_name)
       end
 
       path = Puppet::FileSystem.pathname(Puppet.settings.which_configuration_file)
+      Puppet::FileSystem.dir_mkpath(path)
       Puppet::FileSystem.touch(path)
```
ekohl, line 175: A thought to go in the other direction: should it return if the file doesn't exist? What's the point in creating an empty file and then reading it?

binford2k: it's not "reading" the file, it's opening it and manipulating it. Which ultimately means either _changing_ the contents or _adding_ content -- but the difference is irrelevant.

ekohl: Ah yes, my bad.

PR #172, opened 2025-08-22
treat windows service accounts as case insensitive

binford2k:
Windows service accounts are *case preserving* but not *case sensitive*.

Before this change, Windows service logon account names were compared directly as case sensitive. When there was a case mismatch, it was treated as a change to the account and the service was restarted. This now normalizes case before comparing, reducing spurious restarts.

### DIFF
--- a/lib/puppet/provider/service/windows.rb
+++ b/lib/puppet/provider/service/windows.rb
@@ -127,7 +127,7 @@ Puppet::Type.type(:service).provide :windows, :parent => :service do
     @normalized_logon_account ||= normalize_logonaccount
     @resource[:logonaccount] = @normalized_logon_account
 
-    insync = @resource[:logonaccount] == current
+    insync = @resource[:logonaccount].casecmp?(current)
     self.logonpassword = @resource[:logonpassword] if insync
     insync
   end


bastelfreak:
@binford2k can you please rebase? Tests should pass afterwards.

lib/puppet/provider/service/windows.rb
```diff
@@ -127,7 +127,7 @@ def logonaccount_insync?(current)
     @normalized_logon_account ||= normalize_logonaccount
     @resource[:logonaccount] = @normalized_logon_account
 
-    insync = @resource[:logonaccount] == current
+    insync = @resource[:logonaccount].downcase == current.downcase
```
ekohl: Perhaps use [casecmp?](https://rubyapi.org/o/string#method-i-casecmp-3F) here? I think it's equivalent because https://rubyapi.org/o/string#method-i-casecmp says it's equal to downcasing both.
```suggestion
    insync = @resource[:logonaccount].casecmp?(current)
```

bastelfreak:
@binford2k now can you rebase to get rid of the merge commit? :)

binford2k:
@bastelfreak you can squash or rebase while merging, yeah?

lib/puppet/provider/service/windows.rb
binford2k: Hmm. This assumes that Windows insensitive string matches perform case folding...... That seems like a reasonable assumption to make for now.

binford2k: Hmm. This assumes that Windows insensitive string matches perform case folding...... That seems like a reasonable assumption to make for now.

ekohl: Reading the docs it suggests downcase does the same

PR #182, opened 2025-08-24
CI: Remove unused rspec-mocks dependency

bastelfreak:


bastelfreak:
ok rspec-mocks is a rspec dependency, we pull it in anyways.

PR #184, opened 2025-08-27
Move openvox-agent vanagon repo into this one

nmburgan:
This integrations the openvox-agent vanagon repo into the openvox repo. This allows us to change the version in exactly one place, and have the vanagon build pick that up (and also the gemspec).

This looks daunting to review, but it's mostly just moving all the files from openvox-agent into packaging, and making some pathing tweaks and action changes to accomodate.


binford2k:
I hate to be an asshole, but is it possible to put this into two commits? One to move the files unchanged and then one with the actual fixes for us to review?

nmburgan:
Done

nmburgan:
Probably need to merge the rubocop files and fix up what's going on with all the tests, but the structure of the thing should be reviewable.

lib/puppet/version.rb
```diff
@@ -8,7 +8,7 @@
 # Raketasks and such to set the version based on the output of `git describe`
 
 module Puppet
-  PUPPETVERSION = '8.22.0'
+  PUPPETVERSION = '8.22.1'
```
smortex: Can we commit this change directly in the main branch (or at least, not have it in the middle of this feature branch, because once merged it will be weird)?

openvox.gemspec
```diff
@@ -1,6 +1,15 @@
+# When installed, this gemspec ends up the specifications dir and we should look for
+# the library at the installed path because Puppet ends up loading it when invoked.
+# When we are building the gem from source, we instead look relative to where this gemspec
+# file lives. This is a little jank, but prevents us from having to update the version in
+# two different places.
+installed_path = '/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/version'
+version_path = File.exist?("#{installed_path}.rb") ? installed_path : 'lib/puppet/version'
+require_relative "#{version_path}"
+
```
smortex, line 9: I do not recall the exact details, but IIRC, this gemspec is installed "as is" while in a regular installation (building a gem with gem build, and installing it), the installed gemspec has been generated from the metadata collected in the gemspec at build time.

I think I tried to do something similar (read the version from version file instead of hardcoding), but was probably missing these bits.  I wonder if/how we can avoid this workaround.

smortex (commented):
Thank you for splitting the commits. I have not run the code and skipped the first commit that copy everything, but these changes looks good to me if we remove the unrelated version bump from there.

openvox.gemspec
binford2k: there's worse too! At _install time_ the gemspec is reparsed into a specification which is what rubygems uses to actually load library files. I'm almost 100% certain that the fact that we don't `gem install` is the reason why that "load every single file in `/lib/*` at startup" bug occurs.

nmburgan: Yep, and the gemspec that is actually laid down in the `specifications` folder is actually called `puppet-<version>.gemspec`.  One of these days we'll have to fix it all up :/

lib/puppet/version.rb
nmburgan: Can do!

nmburgan:
Closing in favor of https://github.com/OpenVoxProject/openvox/pull/186

PR #186, opened 2025-08-29
Merge openvox-agent vanagon repo

austb:
Merge with `git merge vanagon-main --allow-unrelated-histories`

The vanagon-main branch is the HEAD of openvox-agent `main` with an
additional commit to move all the content to the packaging subdirectory.


bastelfreak (approved):
I won't review all changed files or commits here.... I am fine with merging this when tests pass.

smortex (approved):
That looks good.  I guess we can only run the tests when this is somewhat merged?  @austb maybe you tested it in your repo (I don't see the action tab there)?

In case it helps other reviewers, I checked changes by updating @austb's remote and running this from the main branch to see all changes except those related to the packaging directory (which is not in the current main branch):

```
git diff -R austb/merge-vanagon -- * .*
```

bastelfreak (approved):

PR #208, opened 2025-09-09
More secure default server setting

binford2k:
Under some fairly obscure conditions, defaulting the server setting to
'puppet' can be a security concern. Because the worst case scenario
involves a user accidentally running `puppet agent -t` on an untrusted
network, this PR removes the default completely when non-root.
Otherwise, it just prints a deprecation warning.

Fixes https://github.com/voxpupuli/security-tracking/issues/22

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -863,7 +863,7 @@ names.
 **Note:** The list of alternate names is locked in when the server's
 certificate is signed. If you need to change the list later, you can't just
 change this setting; you also need to regenerate the certificate. For more
-information on that process, see the 
+information on that process, see the
 [cert regen docs](https://puppet.com/docs/puppet/latest/ssl_regenerate_certificates.html).
 
 To see all the alternate names your servers are using, log into your CA server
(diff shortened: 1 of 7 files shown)


binford2k:
fwiw, I'm not actually happy with this solution. It means that in a completely serverless workflow, you'll have spurious warnings in the logs like such. I'll try to find a better way later, after we decide if this is a good idea or not.

```
$ sudo puppet apply /tmp/site.pp
Warning: OpenVox will not default to `server=puppet` as of version 9.0. Please update your configuration appropriately.
   (location: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/defaults.rb:1659:in `block in initialize_default_settings!')
Notice: Compiled catalog for lilith.local in environment production in 0.00 seconds
...
```

binford2k:
I moved the check and warnings to the http service so it only screams if you actually use it. This avoids, for example, the spurious deprecation message when running `puppet apply`, or even `puppet config set` to set it in the first place!

binford2k:
if people approve of this approach, I'm happy to get started on fixing the spec tests.

bastelfreak:
@binford2k yes this makes sense to me

binford2k:
superseded by #536

PR #210, opened 2025-09-10
Add jruby-9.4.14.0 to CI matrix

bastelfreak:


.github/workflows/tests.yaml
```diff
@@ -48,6 +48,7 @@ jobs:
           - {os: ubuntu-24.04, ruby: 'jruby-9.4'}
           - {os: ubuntu-24.04, ruby: 'jruby-9.4.8.0'}
           - {os: ubuntu-24.04, ruby: 'jruby-9.4.13.0'}
+          - {os: ubuntu-24.04, ruby: 'jruby-9.4.14.0'}
```
nmburgan, line 49: Should we remove .13?

nmburgan: Also, I think `jruby-9.4` automatically picks the latest version. Do we need a specific version here to keep up with latest?

PR #217, opened 2025-09-24
Use cross-compilation for macos-all-x86_64

nmburgan:
Now that we have cross-compilation working, remove the changes made when we were using Rosetta.


bastelfreak (approved):

Sharpie:
Marking as draft due to the `DO NOT MERGE` label.

PR #225, opened 2025-10-06
Introduce the Linux runmode

ekohl:
This builds on the previous work to define all paths in the RunMode class and has a few interesting aspects.

The parts that make it Linux specific is that it prefers [systemd environment variables](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=) to determine locations. When those aren't available, it instead uses hardcoded paths for root and follows the [XDG specification](https://specifications.freedesktop.org/basedir-spec/latest/) for non-root.

This is a draft because it's more of a proposal. It has some implications that may or may not be good. For example, is it a good thing for all of them to be changeable through env vars? It can change the behavior on whether it's a service or a CLI tool.

It's porting over the work I started in https://github.com/puppetlabs/puppet/pull/8636. It's simpler because I no longer rely on `install.rb` in Fedora's packaging that I'm working on.

### DIFF
--- a/lib/puppet/util/run_mode.rb
+++ b/lib/puppet/util/run_mode.rb
@@ -160,5 +160,85 @@ module Puppet
         [ENV.fetch('ALLUSERSPROFILE', nil), "PuppetLabs"] + extra
       end
     end
+
+    # A Linux runmode, using systemd, FHS and XDG standards
+    #
+    # This first attempts systemd environment variables. If those don't exist
+    # it falls back to using hardcoded directories for root and XDG directories
+    # for non-root users. XDG describes various environment variables with
+    # recommended fallbacks.
+    #
+    # @see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=
+    # @see https://specifications.freedesktop.org/basedir-spec/latest/
+    class LinuxRunMode < RunMode
+      def conf_dir
+        ENV.fetch('CONFIGURATION_DIRECTORY') do
+          config_home = which_dir("/etc", ENV.fetch("XDG_CONFIG_HOME", "~/.config"))
+          File.join(config_home, packaging_name)
+        end
+      end
+
+      def code_dir
+        File.join(conf_dir, 'code')
+      end
+
+      def var_dir
+        ENV.fetch('STATE_DIRECTORY') do
+          data_home = which_dir("/var/lib", ENV.fetch("XDG_DATA_HOME", "~/.local/share"))
+          File.join(data_home, packaging_name)
+        end
+      end
+
+      def cache_directory
+        ENV.fetch('CACHE_DIRECTORY') do
+          cache_home = which_dir("/var/cache", ENV.fetch("XDG_CACHE_HOME", "~/.cache"))
+          File.join(cache_home, packaging_name)
+        end
+      end
+
+      def public_dir
+        File.join(cache_directory, 'public')
+      end
+
+      def run_dir
+        ENV.fetch('RUNTIME_DIRECTORY') do
+          runtime_dir = which_dir("/run", ENV.fetch("XDG_RUNTIME_DIR") { File.join('/run', 'user', ::Etc.getpwuid.uid) })
+          File.join(runtime_dir, packaging_name)
+        end
+      end
+
+      def log_dir
+        ENV.fetch('LOGS_DIRECTORY') do
+          which_dir(File.join('/var', 'log', packaging_name),
+                    File.join(ENV.fetch("XDG_STATE_HOME", "~/.local/state"), packaging_name, "logs"))
+        end
+      end
+
+      def pkg_config_path
+        # automatically picked up
+      end
+
+      def gem_cmd
+        '/usr/bin/gem'
+      end
+
+      def data_dir
+        File.join('/usr', 'share', packaging_name)
+      end
+
+      def common_module_dir
+        File.join(data_dir, 'modules')
+      end
+
+      def vendor_module_dir
+        File.join(data_dir, 'vendor_modules')
+      end
+
+      private
+
+      def packaging_name
+        'puppet'
+      end
+    end
   end
 end
--- a/lib/puppet/util/run_mode.rb
+++ b/lib/puppet/util/run_mode.rb
@@ -15,6 +15,8 @@ module Puppet
         @run_modes ||= {}
         if Puppet::Util::Platform.windows?
           @run_modes[name] ||= WindowsRunMode.new(name)
+        elsif Puppet::Util::Platform.linux?
+          @run_modes[name] ||= LinuxRunMode.new(name)
         else
           @run_modes[name] ||= UnixRunMode.new(name)
         end
--- a/lib/puppet/util/platform.rb
+++ b/lib/puppet/util/platform.rb
@@ -21,6 +21,11 @@ module Puppet
       end
       module_function :solaris?
 
+      def linux?
+        RUBY_PLATFORM.include?('linux')
+      end
+      module_function :linux?
+
       def default_paths
         return [] if windows?
 


lib/puppet/util/run_mode.rb
```diff
@@ -160,5 +162,85 @@ def windows_common_base(*extra)
...
+      end
+
+      def vendor_module_dir
+        File.join(data_dir, 'vendor_modules')
+      end
+
+      private
+
+      def packaging_name
+        'puppet'
```
ekohl, line 242: This is to make it easy to switch to OpenVox paths. Perhaps it should be defined in the base class so all packaging can be switched over in 1 go.

lib/puppet/util/run_mode.rb
```diff
@@ -160,5 +162,85 @@ def windows_common_base(*extra)
...
+    class LinuxRunMode < RunMode
+      def conf_dir
+        ENV.fetch('CONFIGURATION_DIRECTORY') do
+          config_home = which_dir("/etc", ENV.fetch("XDG_CONFIG_HOME", "~/.config"))
+          File.join(config_home, packaging_name)
+        end
+      end
+
+      def code_dir
+        File.join(conf_dir, 'code')
```
ekohl, line 184: Does it make sense for non-root users to ignore the `/etc` code directory?

ekohl (commented):
@anarcat and @smortex were involved in the original PR and may have opinions.

lib/puppet/util/run_mode.rb
```diff
@@ -160,5 +162,85 @@ def windows_common_base(*extra)
...
+    # A Linux runmode, using systemd, FHS and XDG standards
+    #
+    # This first attempts systemd environment variables. If those don't exist
+    # it falls back to using hardcoded directories for root and XDG directories
+    # for non-root users. XDG describes various environment variables with
+    # recommended fallbacks.
+    #
+    # @see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=
+    # @see https://specifications.freedesktop.org/basedir-spec/latest/
+    class LinuxRunMode < RunMode
```
smortex, line 175: The `UnixRunMode` is puppetlabs-centric (the `WindowsRunMode` too).  I wonder if we should rather make a hierarchy of classes where we override functions as needed.


A `SystemdRunMode` would basically be a specialization of a `LinuxRunMode` class, where functions call `super` unless the ad-hoc env variable is set.

I also think that the AIO RunMode should be a specialization a of more generic class rather than a base class like it is today, basically some tree of classes like:

* `RunMode`
  * `PosixRunMode`
    * `FreebsdRunMode`
    * `LinuxRunMode`
      * `AIOLinuxRunMode` (what is called `UnixRunMode` today)
      * `SystemdRunMode`
  * `WindowsRunMode`

I am not sure about what levels of hierarchy are required, I put FreeBSD there because I know I would benefit from it.  If it help, here is the special config we set in the generated puppet.conf on FreeBSD so that the various bits are in what we consider the "right" place (where $ETCDIR is `/usr/local/etc/puppet`): https://github.com/freebsd/freebsd-ports/blob/main/sysutils/openvox-agent8/Makefile#L116

lib/puppet/util/run_mode.rb
smortex: I don't think so: when compiling catalogs, doesn't the ccode run as the puppet user?

lib/puppet/util/run_mode.rb
smortex: +1 for having it a single time.

lib/puppet/util/run_mode.rb
```diff
@@ -15,6 +15,8 @@ def self.[](name)
         @run_modes ||= {}
         if Puppet::Util::Platform.windows?
           @run_modes[name] ||= WindowsRunMode.new(name)
+        elsif Puppet::Util::Platform.linux?
+          @run_modes[name] ||= LinuxRunMode.new(name)
         else
           @run_modes[name] ||= UnixRunMode.new(name)
         end
```
smortex, line 22: Multiple run modes will require further logic here de determine which one to prefer.  For some operating systems, a simple test of the os is enough, but for Linux it will probably need something smarted, e.g. detect if AIO or not (to run in AIO RunMode), test if parent PID is 1 or not (for SystemdRunMode), etc.

ekohl: This may be more of an OpenVox 9 question, but do we continue to push the AIO model with paths in `/opt` at all? If so, should the patching patch this class to achieve that instead?

lib/puppet/util/run_mode.rb
ekohl: I was indeed thinking we need something like that hierarchy. I think the splitting up of `LinuxRunMode` may not be needed at all because I'd suggest to drop the specific AIO paths. Only use normal paths. The application code could still be shipped in `/opt` if desired.

anarcat:
for what it is worth, i have no recollection of the original PR nor do i quite understand what this is about. :)

lib/puppet/util/run_mode.rb
anarcat: perhaps .config should be prioritized over /etc, at least when running as a non-root user, though? is that the case here?

at least that's the way it works now, i think...

lib/puppet/util/run_mode.rb
anarcat: hmm... i'm not familiar with this, but we have `UnixRunMode` up there which we seemed to be using before this patch, what is it if not `PosixRunMode` here?

or, to put this another way, why don't we treat all `UnixRunMode` as following the XDG standard? i *think* it's mostly compatible with FreeBSD, for example, except for the `/usr/local` bits

anarcat (approved):
i see. then i think this looks good to me, in general... as i said in the original PR (apparently!) it seems generally sane to me...

lib/puppet/util/run_mode.rb
ekohl: > I don't think so: when compiling catalogs, doesn't the ccode run as the puppet user?

No, the agent runs as root.

> perhaps .config should be prioritized over /etc, at least when running as a non-root user, though? is that the case here?

That should indeed be the case, but today if you install OpenVox AIO packages you get an empty configuration file (only comments) so the built in defaults matter.

lib/puppet/util/run_mode.rb
ekohl: Is XDG a POSIX standard? I don't mind applying that to all platforms.

But the `/usr/local` bit vs `/` is one I have a hard time unifying.

anarcat: XDG is, obviously, not in POSIX.

but then if you derive Linux from POSIX, you're kind of kidding yourself as well... Linux is not, strictly speaking, POSIX either: https://en.wikipedia.org/wiki/POSIX#POSIX-oriented_operating_systems

but we're splitting hairs here: i was mostly wondering where the UnixRunMode fits in your class hierarchy there, i overlooked the AIOLinuxRunMode bits there...

smortex: I agree that XDG can be handled quite low in the hierarchy: it is indeed available on FreeBSD.

ekohl:
Trival rebase, no actual code changes.

lib/puppet/util/run_mode.rb
```diff
@@ -160,5 +162,85 @@ def windows_common_base(*extra)
...
+    # it falls back to using hardcoded directories for root and XDG directories
+    # for non-root users. XDG describes various environment variables with
+    # recommended fallbacks.
+    #
+    # @see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=
+    # @see https://specifications.freedesktop.org/basedir-spec/latest/
+    class LinuxRunMode < RunMode
+      def conf_dir
+        ENV.fetch('CONFIGURATION_DIRECTORY') do
+          config_home = which_dir("/etc", ENV.fetch("XDG_CONFIG_HOME", "~/.config"))
```
MirandaStreeter, line 178: Should we not first check for an existing `~/.puppetlabs` directory, for those migrating or upgrading? And if not present then define the XDG Spec dirs?

MirandaStreeter:
If, for example, `$XDG_CONFIG_HOME` is unset then instead of hardcoding `~/.config/` we should be referencing `$HOME/.config`, per the spec.

lib/puppet/util/run_mode.rb
ekohl: I think that depends on how we want to do it in the migration. For root mode I'd leverage system packaging (RPM / Debian) to migrate, but for non-root I didn't really think about it yet. Is it OK to migrate it on application start up for non-root?

With OpenVox 9 coming up I think it's a good chance to do some breaking changes.

ekohl:
> If, for example, `$XDG_CONFIG_HOME` is unset then instead of hardcoding `~/.config/` we should be referencing `$HOME/.config`, per the spec.

Isn't `~/.config` equivalent to `$HOME/.config`? Reading https://rubyapi.org/o/file#method-c-expand_path
> The given pathname may start with a “~”, which expands to the process owner’s home directory (the environment variable HOME must be set correctly). “~user” expands to the named user’s home directory.

MirandaStreeter:
> Isn't `~/.config` equivalent to `$HOME/.config`? Reading https://rubyapi.org/o/file#method-c-expand_path

Ah, you're right. Apologies, I was getting my wires crossed. Different shells (bash/zsh/csh) can have differing behavior (pulling from `/etc/passwd` instead of referencing `$HOME`). But it looks like the behavior is consistent within ruby, so you're correct, tilde expansion should be fine.

lib/puppet/util/run_mode.rb
MirandaStreeter: From my perspective migrations are a great way to increase the surface area for bugs and corner cases. Perhaps my support-oriented background has given me a more conservative stance, but I don't see any issue with checking for traditional dotfolders and switching to `UnixRunMode` if present. And if one isn't present, continue on with the XDG spec.

MirandaStreeter: And from another angle, if we were aiming for a breaking change we'd have excised `UnixRunMode` entirely.

MirandaStreeter: At the very least, I'm certain there are users out there who decide to muck about with their configs in custom, non-standard ways without referencing `confdir` and such. For that reason displaying a warning when it detects the traditional dotfolder would wake people up to the fact they should check for bespoke config management before we attempt to migrate.

One easy example off the top of my head is laying down some `.conf` files via cloudinit.

A warning would also give us and users some breathing room to make certain there isn't anything we fundamentally broke when switching to the XDG Spec (which don't get me wrong, I totally agree we should move towards).

But, again, this is the support engineer in me talking. :)

bastelfreak:
@ekohl can you please rebase and take a look at the CI?

PR #229, opened 2025-10-20
Perform selabel lookup during file resource synchronization

eciii:
The default values of the SELinux-related properties of a file resource are calculated at the beginning of the catalog application phase, before the catalog resources are actually synchronized. However, by the time the file resource is synchronized, these default values might be outdated due to new policy rules being added in previous parts of the catalog.

This behavior can be easily reproduced in Rocky Linux 9 and probably other RHEL-like distros (maybe even Fedora):

```
$ git checkout main
[...]

$ bundle exec puppet apply -e "package { 'grafana': ensure => present } file { '/var/lib/grafana/test': ensure => file }"
Notice: Compiled catalog for packer-base-rocky9.afdata.local in environment production in 0.37 seconds
Notice: /Stage[main]/Main/Package[grafana]/ensure: created
Notice: /Stage[main]/Main/File[/var/lib/grafana/test]/ensure: created
Notice: Applied catalog in 27.52 seconds

$ ls -lhAZ /var/lib/grafana/
total 0
-rw-r-----. 1 grafana grafana system_u:object_r:grafana_var_lib_t:s0  0 Oct 16 16:55 grafana.db
drwxr-xr-x. 2 grafana grafana system_u:object_r:grafana_var_lib_t:s0 40 Oct 20 14:47 plugins
-rw-r--r--. 1 root    root    system_u:object_r:var_lib_t:s0          0 Oct 20 14:47 test
```

The `test` file is created with the wrong SELinux context because:

- The default value of the `seltype` property is calculated before the `grafana` package is installed.
- The `grafana` package is then installed, pulling in the `grafana-selinux` package, which in turn creates the `grafana_var_lib_t` type.
- The `test` file is created using an outdated SELinux type.

This PR fixes the described behavior:

```
$ rm -f /var/lib/grafana/test && dnf remove -y grafana
[...]

$ git checkout fix
[...]

$ bundle exec puppet apply -e "package { 'grafana': ensure => present } file { '/var/lib/grafana/test': ensure => file }"
Notice: Compiled catalog for packer-base-rocky9.afdata.local in environment production in 0.38 seconds
Notice: /Stage[main]/Main/Package[grafana]/ensure: created
Notice: /Stage[main]/Main/File[/var/lib/grafana/test]/ensure: created
Notice: Applied catalog in 28.41 seconds

$ ls -lhAZ /var/lib/grafana/
total 0
-rw-r-----. 1 grafana grafana system_u:object_r:grafana_var_lib_t:s0  0 Oct 16 16:55 grafana.db
drwxr-xr-x. 2 grafana grafana system_u:object_r:grafana_var_lib_t:s0 40 Oct 20 14:49 plugins
-rw-r--r--. 1 root    root    system_u:object_r:grafana_var_lib_t:s0  0 Oct 20 14:49 test
```

This PR also adjusts the corresponding unit tests accordingly. As far as I cat tell, all the relevant tests pass:

```
$ SPEC=spec/unit/type/file/selinux_spec.rb bundle exec rake spec
rspec spec/unit/type/file/selinux_spec.rb
Run options: exclude {benchmark: true}
....................................

Finished in 0.37257 seconds (files took 0.93635 seconds to load)
36 examples, 0 failures
```

More details about the actual changes can be found in the commit description.

### DIFF
--- a/spec/unit/resource/catalog_spec.rb
+++ b/spec/unit/resource/catalog_spec.rb
@@ -21,6 +21,11 @@ describe Puppet::Resource::Catalog, "when compiling" do
   # audit only resources are unmanaged
   # as are resources without properties with should values
   it "should write its managed resources' types, namevars" do
+    if Puppet.features.selinux?
+      selinux = class_double('selinux', is_selinux_enabled: 0)
+      stub_const('Selinux', selinux)
+    end
+
     catalog = Puppet::Resource::Catalog.new("host")
 
     resourcefile = tmpfile('resourcefile')
--- a/spec/unit/transaction/resource_harness_spec.rb
+++ b/spec/unit/transaction/resource_harness_spec.rb
@@ -29,6 +29,11 @@ describe Puppet::Transaction::ResourceHarness do
 
   describe "when evaluating a resource" do
     it "produces a resource state that describes what happened with the resource" do
+      if Puppet.features.selinux?
+        selinux = class_double('selinux', is_selinux_enabled: 0)
+        stub_const('Selinux', selinux)
+      end
+
       status = @harness.evaluate(@resource)
 
       expect(status.resource).to eq(@resource.ref)


lib/puppet/type/file/selcontext.rb
```diff
@@ -104,7 +106,7 @@ def sync
       enabled."
 
     @event = :file_changed
-    defaultto { retrieve_default_context(:seluser) }
+    defaultto { Puppet::Util::Platform.windows? || @resource[:selinux_ignore_defaults] == :true ? nil : :lookup }
```
binford2k: I don't like repeating this line. And it's too complex to be a one-liner.

I don't think it's correct either. Previously on Windows, this would be `nil` and now it's `true`.

eciii:
@bastelfreak @binford2k I was able to get this commit into good shape, at least in my opinion. I made some small changes with respect with the previous attempt. In particular the `skip_property?` method is introduced to simplify the one-liner that calculates the default value (as Ben suggested).

- I adjusted some tests that were failing

I also ran the full test suite locally on Ubuntu 24.04 LTS and on CentOS Stream 10. This allowed me to catch one bug in the actual proposed solution (catch `nil` returns of `retrieve_default_context` in the `insync?` method) but also in the tests themselves. In particular:

- There was a problem with a test that checks if the `selabel_handle` is closed during transaction teardown. I adjusted this test so that it works but I think the code can be simplified in such a way that this test is made irrelevant. I'll try to put this in a separate PR later.
- Two other tests were failing in CentOS Stream 10 because the test suite runs with SELinux is actually enabled there. The easiest solution was to explicitly disable SELinux for those particular tests. These tests don't fail in GitHub because no RHEL-like distro is used there.

I'll fix the rubocop issue so that the GitHub tests are successful.

eciii:
So there is no one that can review this PR? It would be a pity if this effort gets wasted.

PR #237, opened 2025-11-01
draft: core(exec): Add attribute to check for idempotency attributes

### LINKED ISSUE #13: [Feature request]: make exec more secure
### Use Case

Using exec resource wthout checking idempotency is a bad pattern, as the exec gets executed every time.
Idempotency is possible by setting one of the following parameters:

- creates
- onlyif
- unless
- refreshonly

OpenVox should let users/developers know that they are using code on bad practice.

There are two was on how to achieve this:

1. a lint plugin which identifies if an exec has the idempotency parameters set (out of scope for this project)
2. let the type check if all desired parameters are set

### Describe the solution you would like

The exec type should check if any of the idempotency attributes in use and print a warning (on the agent side) to inform users about this bad practice.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

tuxmea:
fixes #13

### DIFF
--- a/lib/puppet/type/exec.rb
+++ b/lib/puppet/type/exec.rb
@@ -21,6 +21,10 @@ module Puppet
         command only when some other resource is changed. (See the notes on refreshing
         below.)
 
+      The `enable_idempotency_check` attribute is used to check if one of the above mentioned
+      attributes has been set. The parameter defaults to `false` and raises an error if
+      enabled and no idempotency attributes has been set.
+
       The state managed by an `exec` resource represents whether the specified command
       _needs to be_ executed during the catalog run. The target state is always that
       the command does not need to be executed. If the initial state is that the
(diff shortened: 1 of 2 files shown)


tuxmea:
I would like to present this change to the Language Steering Committee.
We can use this PR as discussion basis.

trefzer:
I see noo additional value to check this on every puppet run. For me this is a typical lint check to warn the developer of his 'fault' but at the end it's his decision. There is not additional added value to warn on every puppet run (I know you can supress...) since the running exec is marked as a changing resource anyway. I think we should invest the compute power for this check better. 

Additional remark about the implementation:
It would probably less resource consuming, if you implement the check on the provider side (executed on any change). So it does not run on every puppet run if idempotency is fullfilled. If idempotency is not given, it warns on every run.

PR #240, opened 2025-11-13
include "local" facts in facts output

### LINKED ISSUE #220: [Bug]: `implementation` is just a local fact

### Describe the Bug

We added an implementation fact, which ended up as a local fact: https://github.com/OpenVoxProject/openvox/pull/63 -> https://github.com/OpenVoxProject/openvox/pull/78 -> https://github.com/OpenVoxProject/openvox/pull/107

the fact works fine when running the agent or apply, but it's missing in `puppet facts show implementation`. This is a bug that needs to be fixed. Because of that, the fact is also missing in facterdb factsets.

### Expected Behavior

`puppet facts show implementation` shows `openvox`

### Steps to Reproduce

* install openvox
* run `puppet facts show implementation && puppet apply -e 'notify { $facts["implementation"]: }'`

### Environment

openvox 8.23.1 on all platforms

### Additional Context

_No response_

### Relevant log output

```shell

```

binford2k:
Some years ago [a bug was filed](https://puppet.atlassian.net/browse/PUP-10717)
noting that `puppet facts show` appended all local facts when called with
one or more specific facts. This was resolved by simply not including
local facts for this specific subcommand.

This had the unfortunate side effect that something like `puppet facts
show clientcert` no longer displayed anything. (note: you could still
run `puppet config print certname` but that takes a LOT of institutional
knowledge to know that those are the same thing.) Worse, when we added
the `implementation` local fact, it also didn't display.

This PR just returns to including the local facts into `puppet fact show`
output, but only the ones requested or when requesting all facts.

Fixes #220

### DIFF
--- a/lib/puppet/node/facts.rb
+++ b/lib/puppet/node/facts.rb
@@ -27,11 +27,12 @@ class Puppet::Node::Facts
 
   attr_accessor :name, :values, :timestamp
 
-  def add_local_facts
-    values["implementation"] = Puppet.implementation
-    values["clientcert"] = Puppet.settings[:certname]
-    values["clientversion"] = Puppet.version.to_s
-    values["clientnoop"] = Puppet.settings[:noop]
+  def add_local_facts(query = [])
+    query = Array(query) # some use cases result in a nil being passed in
+    values["implementation"] = Puppet.implementation if query.empty? or query.include? 'implementation'
+    values["clientcert"] = Puppet.settings[:certname] if query.empty? or query.include? 'clientcert'
+    values["clientversion"] = Puppet.version.to_s if query.empty? or query.include? 'clientversion'
+    values["clientnoop"] = Puppet.settings[:noop] if query.empty? or query.include? 'clientnoop'
   end
 
   def initialize(name, values = {})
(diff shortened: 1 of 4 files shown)


lib/puppet/node/facts.rb
```diff
@@ -27,11 +27,12 @@ def save(instance, key = nil, options = {})
 
   attr_accessor :name, :values, :timestamp
 
-  def add_local_facts
-    values["implementation"] = Puppet.implementation
-    values["clientcert"] = Puppet.settings[:certname]
-    values["clientversion"] = Puppet.version.to_s
-    values["clientnoop"] = Puppet.settings[:noop]
+  def add_local_facts(query = [])
+    query = Array(query) # some use cases result in a nil being passed in
```
nmburgan, line 31: Very slightly less boilerplate, but totally fine as-is too.
```suggestion
  def add_local_facts(query)
    query ||= [] # some use cases result in a nil being passed in
```

binford2k: I cast to Array because I'm not 💯 sure that it won't ever get something like a single string.

bastelfreak (approved):

PR #242, opened 2025-11-19
Loosen some gem dependencies for new versions and update minitar usage

nmburgan:
This loosens some dependencies so we can use the latest version. This
includes minitar, which includes a namespace change, so this make the
appropriate changes to use the latest minitar.

### DIFF
--- a/packaging/setup.ps1
+++ b/packaging/setup.ps1
@@ -6,4 +6,4 @@ cmd /c "C:\wix314.exe -quiet"
 $url="https://cygwin.com/setup-x86_64.exe"
 $dest="C:\setup-x86_64.exe"
 Invoke-WebRequest -Uri $url -OutFile $dest
-cmd /c "C:\setup-x86_64.exe -s https://cygwin.osuosl.org -q -P ruby,ruby-devel,gcc-core,make,git,libyaml-devel"
+cmd /c "C:\setup-x86_64.exe -s https://cygwin.osuosl.org -q -P ruby=3.2.2-2,ruby-devel=3.2.2-2,gcc-core,make,git,libyaml-devel"


openvox.gemspec
```diff
@@ -29,8 +29,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
   spec.summary = "OpenVox, a community implementation of Puppet -- an automated configuration management tool"
   spec.specification_version = 4
-  spec.add_runtime_dependency('base64', '>= 0.1', '< 0.4')
-  spec.add_runtime_dependency('benchmark', '>= 0.3', '< 0.6')
+  spec.add_runtime_dependency('base64', '>= 0.1', '< 1')
+  spec.add_runtime_dependency('benchmark', '>= 0.3', '< 1')
```
bastelfreak: I did those very strict version constraints because the rube devs refuse to do a proper 1.0.0 release an honour semver. They said the different 0.X releases can contain breaking changes.

openvox.gemspec
```diff
@@ -39,19 +39,19 @@ Gem::Specification.new do |spec|
...
-    spec.add_runtime_dependency('CFPropertyList', ['>= 3.0.6', '< 4'])
+    spec.add_runtime_dependency('CFPropertyList', ['>= 3.0.6', '< 5'])
   end
 
   if platform == 'x64-mingw32' || platform == 'x86-mingw32'
     # ffi 1.16.0 - 1.16.2 are broken on Windows
-    spec.add_runtime_dependency('ffi', '>= 1.15.5', '< 1.17.0', '!= 1.16.0', '!= 1.16.1', '!= 1.16.2')
-    spec.add_runtime_dependency('minitar', '~> 0.9')
+    spec.add_runtime_dependency('ffi', '>= 1.15.5', '< 2', '!= 1.16.0', '!= 1.16.1', '!= 1.16.2')
+    spec.add_runtime_dependency('minitar', '~> 1.0')
```
bastelfreak, line 55: There was a reason why I didn't update it in the past, but I forgot :( I think the minitar code in openvox has to be adjusted and that's only used on Windows so it wasn't that obvious?🧐

openvox.gemspec
nmburgan: Ugh, really? Don't make breaking changes in minor versions, people.  I'll make them more restrictive again. We're just going to have to be careful that the runtime doesn't automatically update beyond the limit here.

openvox.gemspec
nmburgan: I think I've got the scope of the changes.  I was going to do a separate PR, but the changes are pretty small, so I think I'll just add them to this one.

openvox.gemspec
nmburgan: Fixed

openvox.gemspec
nmburgan: Added those changes.

PR #245, opened 2025-11-21
Enhance platform detection in windows? method

nmburgan:
Update windows? method to improve platform detection. The previous method was not reliable when attempting to run tests via the `bundle exec` from a Cygwin shell. Ruby has largely moved to looking at RUBY_PLATFORM instead anyway.


bastelfreak (approved):
To me this looks like the better implementation, but I am really not a windows expert.

PR #261, opened 2025-12-05
Add dbus feature

ananace:
Should've submitted this earlier, but better late than never.


binford2k (approved):

bastelfreak:
Should we also add the gem to the the gemspec?

nmburgan:
I imagine the presence of the feature would need to be gated on if the gem was present, if we don't add it to the gemspec.

PR #292, opened 2025-12-28
Generate the reference from the runmode

ekohl:
This avoids duplicating the values with the actual logic.

The motivation is to fully centralize all OS specific paths in `RunMode`. If packaging decides to change it (like in https://github.com/OpenVoxProject/openvox/pull/225) then it is limited to a single file.


bastelfreak (approved):

ekohl:
I think this is now ready.

ekohl:
Windows is still failing, but I don't really have a good time testing it locally. Anyone can help?

PR #293, opened 2025-12-28
Drop multi_json support

ekohl:
These days the built in JSON is fast enough and supporting other implementations can only complicate matters. We can also rely on JRuby to have a correct implementation since the same json gem is also built for Java and included in JRuby.

Note I made an assumption about JRuby because https://rubygems.org/gems/json also lists a Java build.

I was triggered to drop this because tests started to fail for me.

### DIFF
--- a/lib/puppet/util/json.rb
+++ b/lib/puppet/util/json.rb
@@ -16,17 +16,7 @@ module Puppet::Util
       end
     end
 
-    begin
-      require 'multi_json'
-      # Force backend detection before attempting to use the library
-      # or load any other JSON libraries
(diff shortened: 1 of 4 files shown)


lib/puppet/util/json.rb
```diff
@@ -16,17 +16,7 @@ def self.build(original_exception, data)
       end
     end
 
-    begin
-      require 'multi_json'
```
bastelfreak, line 20: mental note: we probably can/have to adjust puppet-runtime or openvox-server to not include multi_json in the packages

jay7x (approved):

PR #295, opened 2026-01-06
Use IO.popen to start a process

ekohl:
RuboCop rightfully complains about using open() to spawn a process. This uses the way simpler IO.popen() method.

I have no idea if this is going to break or not and whether it's an API breakage.

### DIFF
--- a/lib/puppet/util/execution.rb
+++ b/lib/puppet/util/execution.rb
@@ -57,31 +57,19 @@ module Puppet::Util::Execution
   # @see Kernel#open for `mode` values
   # @api public
   def self.execpipe(command, failonfail = true)
-    # Paste together an array with spaces.  We used to paste directly
-    # together, no spaces, which made for odd invocations; the user had to
-    # include whitespace between arguments.
-    #
-    # Having two spaces is really not a big drama, since this passes to the
-    # shell anyhow, while no spaces makes for a small developer cost every
-    # time this is invoked. --daniel 2012-02-13
-    command_str = command.respond_to?(:join) ? command.join(' ') : command
-
     if respond_to? :debug
-      debug "Executing '#{command_str}'"
+      debug "Executing #{command.inspect}"
     else
-      Puppet.debug { "Executing '#{command_str}'" }
+      Puppet.debug { "Executing #{command.inspect}" }
     end
 
-    # force the run of the command with
-    # the user/system locale to "C" (via environment variables LANG and LC_*)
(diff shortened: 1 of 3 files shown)


lib/puppet/util/execution.rb
```diff
@@ -57,31 +57,15 @@ def initialize(value, exitstatus)
...
-    # the user/system locale to "C" (via environment variables LANG and LC_*)
-    # it enables to have non localized output for some commands and therefore
-    # a predictable output
-    english_env = ENV.to_hash.merge({ 'LANG' => 'C', 'LC_ALL' => 'C' })
-    output = Puppet::Util.withenv(english_env) do
-      # We are intentionally using 'pipe' with open to launch a process
-      open("| #{command_str} 2>&1") do |pipe| # rubocop:disable Security/Open
-        yield pipe
-      end
+    english_env = ENV.merge({ 'LANG' => 'C', 'LC_ALL' => 'C' })
```
corporate-gadfly: I don't know ruby but should that be?
```suggestion
    english_env = ENV.to_hash.merge({ 'LANG' => 'C', 'LC_ALL' => 'C' })
```

corporate-gadfly (changes requested):

lib/puppet/util/execution.rb
ekohl: Turns out `ENV` is actually an `Object` and not a `Hash`. It just looks like one.

I've now simplified it to just be the overrides because the default behavior of spawning a process is to inherit all the other vars. So I can't think of why the whole merging was needed to begin with.

lib/puppet/util/execution.rb
```diff
@@ -57,31 +57,19 @@ def initialize(value, exitstatus)
...
+    begin
+      output = IO.popen({ 'LANG' => 'C', 'LC_ALL' => 'C' }, command, err: :out) do |pipe|
         yield pipe
       end
+    rescue Errno::ENOENT => e
+      output = e.message
+      raise Puppet::ExecutionFailure, output if failonfail
     end
 
     if failonfail && exitstatus != 0
```
ekohl, line 75: Note to self: I don't know if `exitstatus` is properly set here. https://rubyapi.org/3.4/o/io#method-c-popen doesn't mention setting `$CHILD_STATUS`, only `$?`.

ekohl:
Rebased to resolve conflicts. With OpenVox 9 coming, perhaps a good time to merge this?

PR #307, opened 2026-01-30
Remove monkey patch to remove daemonize

ekohl:
ActiveSupport 3.0.0 dropped this method, which dates back to 2010. It's safe to assume nobody uses such an old version anymore.


ekohl:
Not since https://github.com/OpenVoxProject/openvox/commit/dbb82ef91465ef6a9faebe2850949c2741a797ca.

PR #316, opened 2026-02-09
unbreak OpenBSD package handling.

buzzdeee:
The stock package handling is totally broken since years in Puppet, and it needed patches. Those were reworked about two years ago by myself.

At that time I tried to get it upstreamed to Puppet: https://github.com/puppetlabs/puppet/pull/9301
but never got a response on it.

Since then, the type/provider for OpenBSD is in the ports tree for Puppet 8, and nowadays for OpenVox.

    pkg.conf is long time gone
    drop versionable
    drop upgradeable
    add support for package branches

versionable and upgradeable don't really make much sense on OpenBSD, and they way overcomplicate the provider. However, it could be readded if someone can come up with a compelling usecase ;)

OpenBSD supports package branches, i.e. for puppet, there's puppet 7 and puppet 8, both conflict, and cant be installed at the same time, but with Puppet current package provider, it was impossible to easily tell which of the branches to use. Only when using a fixed "version", or "latest", but that was still quite clumsy, and either had to update the version any time in Hiera when updating puppet, or when using latest, you couldn't choose to install Puppet 7.

There are other packages that have branches, i.e. gimp, which has a stable, and a snapshot branch. They don't conflict with each other, so can be installed in parallel. Puppet with the current package provider would be unable to handle that.

To add support for branches, a new property won't work for those packages with branches that can be installed in parallel, i.e. install both versions of Gimp via Puppet.

The way around it is to specify the branch suffxed to the package name with a leading %

so similarly I would run

```
pkg_add gimp%stable
pkg_add gimp%snapshot
```

I would now use puppet alike:

```
package { "gimp%stable": ensure => "present" }
package { "gimp%snapshot": ensure => "present" }
```

to manage both packages.

Normal packages without branches, would install just normal, for example:

```
package { "gnustep-desktop": ensure => "present" }
```

### DIFF
--- a/lib/puppet/provider/package/openbsd.rb
+++ b/lib/puppet/provider/package/openbsd.rb
@@ -15,8 +15,8 @@ Puppet::Type.type(:package).provide :openbsd, :parent => Puppet::Provider::Packa
     These options should be specified as an array where each element is either a
      string or a hash."
 
-  commands :pkgadd => "pkg_add",
-           :pkginfo => "pkg_info",
+  commands :pkginfo => "pkg_info",
+           :pkgadd => "pkg_add",
            :pkgdelete => "pkg_delete"
 
   defaultfor 'os.name' => :openbsd
@@ -38,7 +38,7 @@ Puppet::Type.type(:package).provide :openbsd, :parent => Puppet::Provider::Packa
 
         # now turn each returned line into a package object
         process.each_line { |line|
-          match = regex.match(line.split("\n")[0])
+          match = regex.match(line.split[0])
           if match
             fields.zip(match.captures) { |field, value|
               hash[field] = value
@@ -88,9 +88,9 @@ Puppet::Type.type(:package).provide :openbsd, :parent => Puppet::Provider::Packa
 
   def get_full_name
     # In case of a real update (i.e., the package already exists) then
-    # pkg_add(8) can handle the flavors. However, if we're actually
+    # pkg_add(1) can handle the flavors. However, if we're actually
     # installing with 'latest', we do need to handle the flavors. This is
-    # done so we can feed pkg_add(8) the full package name to install to
+    # done so we can feed pkg_add(1) the full package name to install to
     # prevent ambiguity.
 
     name_branch_regex = /^(\S*)(%\w*)$/
--- a/spec/unit/provider/package/openbsd_spec.rb
+++ b/spec/unit/provider/package/openbsd_spec.rb
@@ -38,7 +38,7 @@ describe Puppet::Type.type(:package).provider(:openbsd) do
       fixture = File.read(my_fixture('pkginfo.list'))
       expect(described_class).to receive(:execpipe).with(%w{/bin/pkg_info -a -z}).and_yield(fixture)
       expect(described_class.instances.map(&:name).sort).to eq(
-        %w{autoconf%2.13 autoconf%2.56 bash postfix%stable puppet%8 zstd}.sort
+        %w{autoconf%2.13 autoconf%2.56 bash postfix%stable openvox%8 zstd}.sort
       )
     end
 
@@ -52,7 +52,7 @@ describe Puppet::Type.type(:package).provider(:openbsd) do
       expect(instances[1]).to eq({:name => 'autoconf%2.56', :flavor => :absent, :branch => '%2.56'})
       expect(instances[2]).to eq({:name => 'bash', :flavor => :absent, :branch => :absent})
       expect(instances[3]).to eq({:name => 'postfix%stable', :flavor => 'ldap', :branch => '%stable'})
-      expect(instances[4]).to eq({:name => 'puppet%8', :flavor => :absent, :branch => '%8'})
+      expect(instances[4]).to eq({:name => 'openvox%8', :flavor => :absent, :branch => '%8'})
       expect(instances[5]).to eq({:name => 'zstd', :flavor => :absent, :branch => :absent})
     end
   end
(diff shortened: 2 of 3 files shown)


lib/puppet/provider/package/openbsd.rb
```diff
@@ -6,22 +6,24 @@
...
+
     This provider supports the `install_options` and `uninstall_options`
     attributes, which allow command-line flags to be passed to pkg_add and pkg_delete.
     These options should be specified as an array where each element is either a
      string or a hash."
 
-  commands :pkginfo => "pkg_info",
-           :pkgadd => "pkg_add",
+  commands :pkgadd => "pkg_add",
+           :pkginfo => "pkg_info",
```
bastelfreak: was there a reason for flipping the order?

lib/puppet/provider/package/openbsd.rb
```diff
@@ -30,20 +32,21 @@ def self.instances
...
-        regex = /^(.*)-(\d[^-]*)-?([\w-]*)(.*)$/
-        fields = [:name, :ensure, :flavor]
+        regex = /^(.*)--([\w-]+)?(%[^w]+)?$/
+        fields = [:name, :flavor, :branch]
         hash = {}
 
         # now turn each returned line into a package object
         process.each_line { |line|
-          match = regex.match(line.split[0])
+          match = regex.match(line.split("\n")[0])
```
bastelfreak: I thought the `.each_line` in line 40 basically does `.split("\n")`? Does that split every do anything here?

bastelfreak:
Ahhh the patch is already used downstream, perfect!

lib/puppet/provider/package/openbsd.rb
klemensn: Looks like churn.

lib/puppet/provider/package/openbsd.rb
```diff
@@ -63,191 +66,71 @@ def self.instances
...
+      self.fail "pkg_add returned: #{output.chomp}"
     end
-
-    Puppet::Util.withenv(e_vars) { pkgadd cmd.flatten.compact }
   end
 
-  def get_full_name(latest = false)
+  def get_full_name
     # In case of a real update (i.e., the package already exists) then
     # pkg_add(8) can handle the flavors. However, if we're actually
```
klemensn: It is pkg_add(1).

spec/unit/provider/package/openbsd_spec.rb
```diff
@@ -46,295 +20,91 @@ def expect_pkgadd_with_env_and_name(source, &block)
...
-      expect(instances.size).to eq(2)
-      expect(instances[0]).to eq({:name => 'bash', :ensure => '3.1.17', :flavor => 'static'})
-      expect(instances[1]).to eq({:name => 'vim',  :ensure => '7.0.42', :flavor => 'no_x11'})
+        :flavor => p.get(:flavor), :branch => p.get(:branch)}}
+      expect(instances.size).to eq(6)
+      expect(instances[0]).to eq({:name => 'autoconf%2.13', :flavor => :absent, :branch => '%2.13'})
+      expect(instances[1]).to eq({:name => 'autoconf%2.56', :flavor => :absent, :branch => '%2.56'})
+      expect(instances[2]).to eq({:name => 'bash', :flavor => :absent, :branch => :absent})
+      expect(instances[3]).to eq({:name => 'postfix%stable', :flavor => 'ldap', :branch => '%stable'})
+      expect(instances[4]).to eq({:name => 'puppet%8', :flavor => :absent, :branch => '%8'})
```
klemensn: Do these tests actually run on OpenBSD and install those packages?

If so, it'd be wise to use `openvox%8` instead as the `sysutils/ruby-puppet` port will removed soon.

spec/unit/provider/package/openbsd_spec.rb
```diff
@@ -46,295 +20,91 @@ def expect_pkgadd_with_env_and_name(source, &block)
...
 
-    it "should lookup the correction version with flavors" do
-      provider.resource[:name] = 'fossil'
-      provider.resource[:flavor] = 'static'
-      output = 'fossil-1.29v0-static simple distributed software configuration management'
-      expect(provider).to receive(:execpipe).with(%w{/bin/pkg_info -I fossil}).and_yield(output)
-      expect(provider.get_full_name).to eq('fossil-1.29v0-static')
+    it "should return the full unversioned package name when installing without a flavor" do
+        provider.resource[:name] = 'puppet'
+        expect(provider.get_full_name).to eq('puppet--')
```
klemensn: Similar nit here:  why not embrace the new project? ;)

klemensn (approved):
This already works fine in my setup.

A few things came up after glancing over the code, but nothing that should prevent this from going in as-is, imho.

One thing I noticed just now is how the provider treats branchless package names for branched packages:

```
$ pkg_info -c openvox
Information for inst:openvox-8.24.2p0

Comment:
community implementation of Puppet


```

If there's only one branch, there's no choice, you get no prompt during installation and pkg_*(1) generally uses the only branch there is, e.g. `pkg_delete openvox` uninstalls above package just fine.

Doing the same with OpenVox gives different results:
```
$ puppet resource package openvox
package { 'openvox':
  ensure   => ['8.24.2'],
  provider => 'gem',
}

$ puppet resource package openvox%8
package { 'openvox%8':
  provider => 'openbsd',
}
```

I'm not sure whether this is a shortcoming or necessity.

It is not a problem per se, imho, but does stick out;  if expected, perhaps a comment/note in some place would be appropiate, stressing the need for `%8`-like branch names in Puppet code.

spec/unit/provider/package/openbsd_spec.rb
buzzdeee: It's based on a fixture file, they don't run on a life system.

buzzdeee:
indeed ;)

Issue #298, opened 2026-01-22
[Feature request]: environment parameter for package resource


TuningYourCode:
### Use Case

Some 3rd party packages require that during installation some environment variables are set.

A common example would be opensearch which requires to have set `OPENSEARCH_INITIAL_ADMIN_PASSWORD` that the package installation does not fail.
https://docs.opensearch.org/latest/install-and-configure/install-opensearch/debian/#install-opensearch-from-a-package

Another example would be percona where you can disable telemetry by setting an environment variable:
https://docs.percona.com/percona-server/innovation-release/telemetry.html#get-expert-help

### Describe the solution you would like

The ideal solution would be if the package resource would support passing environment variables.

```
package { 'opensearch':
   ensure => 'present',
   environment => [
      "OPENSEARCH_INITIAL_ADMIN_PASSWORD=${password}",
   ],
}
```

Implementation should most likely be the same like in the `exec` resource where it's already possible to pass environment variables:
https://help.puppet.com/core/8/Content/PuppetCore/Markdown/exec.htm?Highlight=exec#exec-attribute-environment

### Describe alternatives you've considered

We currently use `exec` as an alternative but it requires quiet some more knowledge and is not plattform independent code.

See: https://github.com/voxpupuli/puppet-opensearch/pull/71

PR #321, opened 2026-02-12
Add environment parameter to the package resource type

C24-AK:
Allow passing environment variables to package management commands, mirroring the exec resource's environment parameter. This enables packages that require environment variables during installation, such as OpenSearch (OPENSEARCH_INITIAL_ADMIN_PASSWORD) or Percona (PERCONA_TELEMETRY_DISABLE).

The implementation uses a dual approach:
- Override execute() in the base Package provider to inject :custom_environment for providers using execute() directly (yum, dnf)
- Add with_environment helper using Puppet::Util.withenv for providers using named command methods (apt, dpkg, rpm, zypper)

Resolves: https://github.com/OpenVoxProject/openvox/issues/298

### DIFF
--- a/lib/puppet/provider/package.rb
+++ b/lib/puppet/provider/package.rb
@@ -51,7 +51,7 @@ class Puppet::Provider::Package < Puppet::Provider
 
     envlist = [envlist] unless envlist.is_a? Array
     envlist.each do |setting|
-      unless (match = /^(\w+)=((.|\n)*)$/.match(setting))
+      unless (match = /\A(\w+)=((.|\n)*)\z/.match(setting))
         warning _("Cannot understand environment setting %{setting}") % { setting: setting.inspect }
         next
       end
--- a/lib/puppet/type/package.rb
+++ b/lib/puppet/type/package.rb
@@ -550,7 +550,7 @@ module Puppet
       validate do |values|
         values = [values] unless values.is_a? Array
         values.each do |value|
-          unless value =~ /\w+=/
+          unless value =~ /\A\w+=/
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/package.rb
```diff
@@ -35,6 +35,73 @@ def validate_source(value)
...
+
+    env_param = resource.parameters[:environment]
+    return env unless env_param
+
+    envlist = env_param.value
+    return env unless envlist
+
+    envlist = [envlist] unless envlist.is_a? Array
+    envlist.each do |setting|
+      unless (match = /^(\w+)=((.|\n)*)$/.match(setting))
```
Sharpie: Anchor regexes with `\A` and `\Z`, start of string and end of string, instead of `^` and `$`, start of line and end of line, as the latter will get thrown off by newline characters.

lib/puppet/type/package.rb
```diff
@@ -533,6 +533,30 @@ def change_to_s(currentvalue, newvalue)
...
+            }
+
+        These variables are applied to all package management commands
+        (install, update, uninstall, purge) executed by the provider.
+      EOT
+
+      validate do |values|
+        values = [values] unless values.is_a? Array
+        values.each do |value|
+          unless value =~ /\w+=/
```
Sharpie: The regex used to validate should be left-anchored to the start of string: `/\A\w+=/`

lib/puppet/provider/package.rb
C24-AK: Pushed in my commit

lib/puppet/type/package.rb
C24-AK: Pushed in my commit

Issue #328, opened 2026-02-18
[Feature request]: file resource should support etag


deric:
### Use Case

File resources fetched from CDN sometimes provide header like `date` which causes unnecessary corrective changes. 

```
$ curl https://repos.influxdata.com/influxdata-archive.key -I
HTTP/2 200 
date: Sat, 07 Feb 2026 21:14:40 GMT
content-type: application/vnd.apple.keynote
content-length: 6198
vary: accept-encoding
cache-control: public, max-age=31536000
etag: "f5ffec8d8d16b43d5e9ac6ad4330c445"
x-cache: MISS
server: cloudflare
cf-ray: 9ca5e5f089e9d9db-FRA
```
Currently `last-modified` header [is supported](https://github.com/OpenVoxProject/openvox/blob/main/lib/puppet/file_serving/http_metadata.rb#L38). 

### Describe the solution you would like

It would be nice to support  e.g. `etag` header that would indicate that the remote content hasn't changed.

### Additional context

See:

- https://github.com/voxpupuli/puppet-telegraf/pull/267
- https://github.com/puppetlabs/puppetlabs-apt/issues/1196

deric:
@bastelfreak I could try that, although I don't have much experience with writing openvox code. Are you ok with the `etag` approach or do you have some better idea how to approach this?

PR #329, opened 2026-02-18
Feature: file etag support

C24-AK:
Support ETag header in HTTP file metadata

When sourcing files from HTTP, use the ETag header as a content
checksum when it contains a recognizable hash (MD5, SHA1, SHA256).
This prevents unnecessary corrective changes on every Puppet run
for CDN-hosted files that return ETag but no Last-Modified header.

Explicit checksum headers (X-Checksum-*, Content-MD5) take priority
over ETag. Weak ETags and non-hash values are ignored gracefully.

Reference Documentation: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag

Resolves: https://github.com/OpenVoxProject/openvox/issues/328

### DIFF
--- a/spec/unit/indirector/catalog/compiler_spec.rb
+++ b/spec/unit/indirector/catalog/compiler_spec.rb
@@ -226,7 +226,7 @@ describe Puppet::Resource::Catalog::Compiler do
       allow(node.environment).to receive(:static_catalogs?).and_return(true)
 
       expect { compiler.find(@request) }.to raise_error Puppet::Error,
-        "Unable to find a common checksum type between agent 'atime.md2' and master '[:sha256, :sha256lite, :md5, :md5lite, :sha1, :sha1lite, :sha512, :sha384, :sha224, :mtime, :ctime, :none]'."
+        "Unable to find a common checksum type between agent 'atime.md2' and master '[:sha256, :sha256lite, :md5, :md5lite, :sha1, :sha1lite, :sha512, :sha384, :sha224, :mtime, :ctime, :none, :etag]'."
     end
 
     it "errors if checksum_type contains no shared checksum types" do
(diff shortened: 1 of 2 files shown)


C24-AK:
@bastelfreak Should be fine now.

bastelfreak (approved):
looks good to me, but I would like to get another set of eyes on this

Sharpie:
A separate parameter would also work. If I'm reading the changeset correctly, it looks like the current implementation auto-activates whenever an `ETag` header is present in the server response.

The problem with `ETag` is that the RFC standards did not define a strict format for the header value. So, having this behavior auto-activate injects a small amount of undefined behavior into all existing file resources --- which is a bit risky.

I think this needs to be opt-in, so that it can be activated only for `file` resources where manifest authors know the behavior of the file server matches the assumptions made by the parsing logic.

Sharpie (approved):
Applied this PR as a patch on top of `openvox-agent` 8.25.0 and tested against a Sinatra webserver:

```ruby
#!/opt/puppetlabs/puppet/bin/ruby
require 'sinatra'
require 'digest/md5'
require 'digest/sha1'
require 'digest/sha2'

MD5 = Digest::MD5.hexdigest("Hello, md5!\n")
SHA1 = Digest::SHA1.hexdigest("Hello, sha1!\n")
SHA2 = Digest::SHA256.hexdigest("Hello, sha256!\n")

get '/md5' do
  [200, {'ETag' => "\"#{MD5}\""}, "Hello, md5!\n"]
end

get '/sha1' do
  [200, {'ETag' => "\"#{SHA1}\""}, "Hello, sha1!\n"]
end

get '/sha256' do
  [200, {'ETag' => "\"#{SHA2}\""}, "Hello, sha256!\n"]
end
```

Everything works as expected 👍

PR #382, opened 2026-04-07
deprecate sha1() Puppet function

silug:
SHA1 is no longer considered safe for new use. Emit a puppet_deprecation_warning on every call to sha1() directing users to the sha256() function from puppetlabs-stdlib.

### Short description
SHA1 is no longer considered safe for new use. Emit a puppet_deprecation_warning on every call to sha1() directing users to the sha256() function from puppetlabs-stdlib.

### Checklist
I have:
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/parser/functions/sha1.rb
+++ b/lib/puppet/parser/functions/sha1.rb
@@ -2,6 +2,14 @@
 
 require 'digest/sha1'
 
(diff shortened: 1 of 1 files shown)


Sharpie:
I'm not sure about this one. SHA1 may no longer be recommended for new use, but it seems likely that config management will involve comparing SHA-1 checksums `$forever`.

The `git` ecosystem is certainly going to coast along on SHA-1 until something truly seismic happens.

PR #419, opened 2026-04-24
Remove filetimeout deprecation warning

silug:
### Short description
The filetimeout setting was deprecated in favor of JRuby timers, but since we're moving away from the JVM the fine-grained control is useful again. Remove the deprecation warning and restore full setting semantics.

Fixes #345.

### Checklist
I have:
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -694,21 +694,8 @@ module Puppet
       :type       => :duration,
       :desc       => "The minimum time to wait between checking for updates in
       configuration files.  This timeout determines how quickly Puppet checks whether
-      a file (such as manifests or puppet.conf) has changed on disk. The default will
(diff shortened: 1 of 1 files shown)


bastelfreak:
I am not quite sure how this option differs from the environment timeout setting. Do we need both options?

silug:
> I am not quite sure how this option differs from the environment timeout setting. Do we need both options?

@bastelfreak My understanding is that `filetimeout` is used for file watches outside the environment (`puppet.conf`, for example).

Sharpie (approved):
As I understand it, the rationale for deprecating this in the first place was to allow Java/Clojure filesystem watching logic to take over.

Given we want to keep our non-Java options open going into next year, un-deprecating is the right move 👍

PR #432, opened 2026-05-07
Notify now honors `withpath` parameter

acesso:
When running Puppet agent with the -t flag, notify resources currently log duplicate messages: one for the explicit message itself, and another for the resource state change (creation/sync). This creates unnecessary log noise, making it harder to spot actual infrastructure changes.

This change modifies the notify resource to always be considered as "existing." By preventing the resource from triggering an out-of-sync event, Puppet will no longer log the redundant state-change notification, resulting in cleaner and more readable execution logs.

The `withpath` parameter now makes sense to be used, since the possible need to actually show the resource path.

### Checklist
I have:
- [ ] read and accepted the [Developer Certificate of Origin](https://github.com/OpenVoxProject/.github/blob/main/DCO.md) document and added a [`Signed-off-by`](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) annotation to each of my commits
- [x] tested this code
- [x] documented the code

### DIFF
--- a/lib/puppet/type/notify.rb
+++ b/lib/puppet/type/notify.rb
@@ -11,7 +11,7 @@ module Puppet
 
     newproperty(:message, :idempotent => false) do
       desc "The message to be sent to the log. Note that the value specified must be a string."
-      def sync
+      def insync?(is)
         message = @sensitive ? 'Sensitive [value redacted]' : should
         case @resource["withpath"]
         when :true
@@ -19,15 +19,11 @@ module Puppet
         else
           Puppet.send(@resource[:loglevel], message)
         end
-        nil
+        true
       end
 
       def retrieve
         :absent
-      end
-
-      def insync?(is)
-        false
       end
 
       defaultto { @resource[:name] }


Sharpie (changes requested):
Apologies for the delay in reviewing this PR. This looks like a good set of changes for OpenVox 9. @acesso could you re-base your pull request against the `main` branch?

That should kick the test suite. The commit in this PR will also need to be signed with a SSH or GPG key.

Sharpie:
Looks like this caused quite a few failures in the testing suite. I applied the patch locally and compared `/opt/puppetlabs/puppet/cache/state/last_run_report.yaml` before and after. Looks like this prevents the `notify` resource from registering a resource event --- which is likely the cause of many of the test failures.

Unfortunately, event creation is pretty intrinsic to how `notify` works. If we did land this change, it would have to be opt-in as I suspect there are quite a few installations out there that are processing reports and looking for the presence of `notify` events as an alert indicator.

Alternately, the `ipcrm-echo` module provides an `echo` resource which is just logging and no event creation: https://forge.puppet.com/modules/ipcrm/echo/

PR #440, opened 2026-05-12
Update ca/crl last_update on HTTP 304 so refresh intervals are respected (#435)

### LINKED ISSUE #435: [Bug]: ca_refresh_interval is not respected — agent refreshes CA on every run when server responds 304

### Describe the Bug

The `ca_refresh_interval` setting is not respected in practice. Agents attempt
  to refresh the CA certificate on every run instead of at the configured
  interval (default: 24 hours), resulting in an unnecessary HTTP request to the
  CA endpoint per run.

  The analogous `crl_refresh_interval` logic contains the same structural issue
  but does not manifest in practice (see "Why CRL is not affected" below).

## Actual Behavior
  The agent attempts to refresh the CA certificate on every agent run. Debug
  logs show "Refreshing CA certificate" / "CA certificate is unmodified"
  appearing on every run, not at the configured interval.

  The mtime on `$certdir/ca.pem` is never updated after the initial download,
  so `needs_refresh?` always evaluates to true.

### Expected Behavior

With `ca_refresh_interval = 1d`, the agent should attempt to refresh the CA
  certificate approximately once per 24 hours.

### Steps to Reproduce

  1. Configure an agent with default `ca_refresh_interval` (1d)
  2. Let the agent run and successfully download the CA
  3. Trigger subsequent agent runs at any cadence
  4. Observe: `puppet agent -t --debug 2>&1 | grep -E "Refreshing CA|unmodified"`
     prints on every run, not once per 24 hours
  5. `stat $(puppet config print localcacert)` shows mtime is not updated

### Environment

Version 8.26.2
Platform Rocky Linux release 9.7 (Blue Onyx)

## Root Cause

  In `lib/puppet/ssl/state_machine.rb`, `NeedCACerts#refresh_ca` (lines 111-135)
  updates `@cert_provider.ca_last_update` only on the success (HTTP 200) path.
  When the server responds 304 Not Modified, `download_ca` raises
  `Puppet::HTTP::ResponseError`, the rescue handler logs "CA certificate is
  unmodified" but does not update `ca_last_update`:

  ```ruby
  def refresh_ca(ssl_ctx, last_update)
    Puppet.info(_("Refreshing CA certificate"))
    next_ctx = [download_ca(ssl_ctx, last_update), true]
    @cert_provider.ca_last_update = Time.now   # only reached on HTTP 200
    next_ctx
  rescue Puppet::HTTP::ResponseError => e
    if e.response.code == 304
      Puppet.info(_("CA certificate is unmodified, using existing CA certificate"))
      # ca_last_update is not updated here
    else
      Puppet.info(_("Failed to refresh CA certificate..."))
    end
    [ssl_ctx, false]
  end
```
  Since `ca_last_update` is backed by the mtime of `ca.pem`
  (`lib/puppet/x509/cert_provider.rb`:[159-173](https://github.com/OpenVoxProject/openvox/blob/0c02b1c0777677eff310c08144c38939db6817d9/lib/puppet/x509/cert_provider.rb#L159-L173)), and neither `save_cacerts` nor
  `ca_last_update=` runs on the 304 path, the mtime stays at whatever it was
  after the initial download. On the next run, `needs_refresh?` sees a
  stale mtime and schedules another refresh — repeating indefinitely.

### Why CRL is Not Affected in Practice

  `NeedCRLs#refresh_crl` (lines [209-233](https://github.com/OpenVoxProject/openvox/blob/0c02b1c0777677eff310c08144c38939db6817d9/lib/puppet/ssl/state_machine.rb#L209-L233)) has the identical structural issue,
  but in typical deployments the server returns HTTP 200 for CRL requests
  often enough that `save_crls` writes the file and bumps the mtime naturally.
  For the CA certificate, which rarely changes, 304 is the steady-state
  response and the bug is visible.

## Suggested Fix

  On HTTP 304, update the last-update timestamp — a 304 is positive
  confirmation that the local copy is current, so the TTL should be reset:

  ```ruby
  rescue Puppet::HTTP::ResponseError => e
    if e.response.code == 304
   Puppet.info(_("CA certificate is unmodified, using existing CA certificate"))
   @cert_provider.ca_last_update = Time.now
    else
   Puppet.info(_("Failed to refresh CA certificate, using existing CA certificate: %{message}") % { message: e.message })
    end
    [ssl_ctx, false]
  end
```

  The same fix applies to `refresh_crl` for correctness, even though the
  symptom is latent there.

 ---

  Disclosure: initial analysis of this issue was done with AI assistance
  (Claude). The code references, reproduction steps, and root cause have been
  verified against the source at the paths cited above.

### Relevant log output

```shell

```

SAY-5:
Fixes #435.

When the CA endpoint responds 304 Not Modified, `refresh_ca` previously logged "CA certificate is unmodified" but never reset `ca_last_update`. Since that timestamp is backed by the mtime of `ca.pem`, and the mtime is only bumped on the 200 path through `save_cacerts`, in steady state (where 304 is the typical response) the mtime stayed pinned to the initial download. As a result `needs_refresh?` evaluated to true on every agent run, and the agent issued a conditional GET to the CA every cycle regardless of the configured `ca_refresh_interval`.

`refresh_crl` has the same structural issue but is latent in practice because CRL endpoints return 200 often enough that `save_crls` bumps the mtime naturally. Fixed for symmetry.

Reproduction (from the issue):
```
puppet agent -t --debug 2>&1 | grep -E 'Refreshing CA|unmodified'
```
Prior behavior: logs print on every run. With this patch: logs print at the configured `ca_refresh_interval` (default 24h).

Regression tests added:
- ``updates the `last_update` time on 304 Not Modified so ca_refresh_interval is respected``
- ``updates the `last_update` time on 304 Not Modified so crl_refresh_interval is respected``

Both fail without the patch and pass with it. Full `spec/unit/ssl/state_machine_spec.rb` is green (99 examples).

### DIFF
--- a/lib/puppet/ssl/state_machine.rb
+++ b/lib/puppet/ssl/state_machine.rb
@@ -121,6 +121,10 @@ class Puppet::SSL::StateMachine
     rescue Puppet::HTTP::ResponseError => e
       if e.response.code == 304
         Puppet.info(_("CA certificate is unmodified, using existing CA certificate"))
+        # 304 confirms the local CA is current, so reset the TTL.
+        # Otherwise ca_last_update (backed by ca.pem mtime) never advances
+        # past the initial download and needs_refresh? triggers on every run.
+        @cert_provider.ca_last_update = Time.now
       else
         Puppet.info(_("Failed to refresh CA certificate, using existing CA certificate: %{message}") % { message: e.message })
       end
@@ -219,6 +223,8 @@ class Puppet::SSL::StateMachine
     rescue Puppet::HTTP::ResponseError => e
       if e.response.code == 304
         Puppet.info(_("CRL is unmodified, using existing CRL"))
+        # 304 confirms the local CRL is current, so reset the TTL.
+        @cert_provider.crl_last_update = Time.now
       else
         Puppet.info(_("Failed to refresh CRL, using existing CRL: %{message}") % { message: e.message })
       end
--- a/spec/unit/ssl/state_machine_spec.rb
+++ b/spec/unit/ssl/state_machine_spec.rb
@@ -495,6 +495,14 @@ describe Puppet::SSL::StateMachine, unless: Puppet::Util::Platform.jruby? do
         state.next_state
       end
 
+      it 'updates the `last_update` time on 304 Not Modified so ca_refresh_interval is respected' do
+        stub_request(:get, %r{puppet-ca/v1/certificate/ca}).to_return(status: 304)
+
+        expect_any_instance_of(Puppet::X509::CertProvider).to receive(:ca_last_update=).with(be_within(60).of(Time.now))
+
+        state.next_state
+      end
+
       it "does not update the `last_update` time when CA refresh fails" do
         stub_request(:get, %r{puppet-ca/v1/certificate/ca}).to_raise(Errno::ECONNREFUSED)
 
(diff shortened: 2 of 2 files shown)


Sharpie (approved):
This looks good to me. The commit will need to be signed with a SSH or GPG key in order for this PR to be mergable.

PR #481, opened 2026-06-09
Fix regex node lookup regression from PUP-11515

### LINKED ISSUE #14: [Bug]: Puppet 8 introduced a regex node definition regression

### Describe the Bug

> We have node definitions of the form node /^thing-(foo|bar)(?:-test)?-(\d+)/, the backreferences in which are known to trigger https://github.com/puppetlabs/puppet/issues/9329 in Puppet 7. We thought we'd take a look at Puppet 8 and see if that behavior has changed. However it seems that this pattern doesn't work at all in Puppet 8:

```
ERROR [qtp437212910-1916] [puppetserver] Puppet Server Error: Cannot find definition Node on node thing-foo-0.example.com
```

See https://github.com/puppetlabs/puppet/issues/9538 for more information

It looks like the regression was introduced in https://github.com/openvox/puppet/commit/5d09d7f19ab7c68e1bc4b0df88dc1231684548b3

### Expected Behavior

regex should work, or it should be documented why it does not work.

### Steps to Reproduce

Create a complex regex node definition

### Environment

Puppet/OpenVox 8.x+

### Additional Context

_No response_

### Relevant log output

```shell

```

corporate-gadfly:
### Short description
The LOOKAROUND_OPERATORS tokens introduced in 5d09d7f are uppercase, but TypeCollection#munge_name() lowercases all lookup keys, so regex nodes with group syntax (e.g. (?:-test)) were stored under a key that could never be found again, producing "Cannot find definition Node".

Fix: use downcased lookaround operators.

Fixes: https://github.com/OpenVoxProject/openvox/issues/14

### Checklist

I have:
- [x] tested this code
- [x] documented the code

### DIFF
--- a/lib/puppet/resource/type.rb
+++ b/lib/puppet/resource/type.rb
@@ -34,13 +34,13 @@ class Puppet::Resource::Type
   EMPTY_ARRAY = [].freeze
 
   LOOKAROUND_OPERATORS = {
-    "(" => 'LP',
-    "?" => "QU",
-    "<" => "LT",
-    ">" => "GT",
-    "!" => "EX",
-    "=" => "EQ",
-    ")" => 'RP'
+    "(" => 'lp',
+    "?" => "qu",
+    "<" => "lt",
+    ">" => "gt",
+    "!" => "ex",
+    "=" => "eq",
+    ")" => 'rp'
   }.freeze
 
   attr_accessor :file, :line, :doc, :code, :parent, :resource_type_collection, :override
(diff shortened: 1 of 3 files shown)


binford2k:
This is a relatively harmless failure. It just means that on Windows / Ruby 3.2 that the generate face will do more work than it has to do. It should only be generating types for _changed_ files, but with this change it's generating for all. It's still generating properly.

I don't have time to keep digging just now, but you might poke at https://github.com/OpenVoxProject/openvox/blob/acd591a4148a687bc01b2be122b9b0311520055d/lib/puppet/generate/type.rb#L47-L53

Either something about this patch changes how the path is calculated or something about this patch changes how mtime works.

The only link I found between node name and the generate types command is that it uses `Puppet::Node::Environment` to identify where to look at files. If you have a Windows node handy, you might see if there's any difference in `Puppet.lookup(:current_environment)` in the context of that test.

corporate-gadfly:
Thankfully, my life is devoid of Windows at the moment :shrug:.

bastelfreak (approved):

corporate-gadfly:
FYI, flaky test failure was fixed by PR #522.

PR #536, opened 2026-07-09
More secure default server setting

corporate-gadfly:
Under some fairly obscure conditions, defaulting the server setting to 'puppet' can be a security concern. Because the worst case scenario involves a user accidentally running `puppet agent -t` on an untrusted network, this PR removes the default completely when non-root. Otherwise, it just prints a deprecation warning.

The check and warnings are generated at the http service so that it only screams if you actually attempt to use it. This avoids, for example, the spurious deprecation message when running `puppet config set` to set this in the first place.

Fixes https://github.com/voxpupuli/security-tracking/issues/22

### Checklist

I have:
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -842,7 +842,7 @@ names.
 **Note:** The list of alternate names is locked in when the server's
 certificate is signed. If you need to change the list later, you can't just
 change this setting; you also need to regenerate the certificate. For more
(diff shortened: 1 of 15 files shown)


binford2k (approved):

JRDavisUF:
Can/should this be tweaked to check the setting of "server" or "server_list"?

PR #550, opened 2026-07-15
Fix ineffective if warnings in discarded case branches

### LINKED ISSUE #542: [Bug]: Sometimes Error "This 'if' statement has no effect", sometimes not

### Describe the Bug

Puppet gives this error `This 'if' statement has no effect. A value was produced and then forgotten (one or more preceding expressions may have the wrong form) (line: 13, column: 7)` with the following code:

```puppet
# class dcm::application::default_apps
# @summary This class contains a list of default applications for each Distribution/OS
# @api private
#
class dcm::application::default_apps () {
  assert_private()

  $os_name = $facts['os']['name']
  case $os_name {
    'Ubuntu', 'SLES': {
      $packages = ['ncdu', 'netcat-openbsd', 'less', 'nano', 'vim', 'mokutil', 'btop']

      if $facts['is_virtual'] and $facts['dmi']['manufacturer'] == 'VMware, Inc.' { # error here
        $packages << 'open-vm-tools'
      }

      if $os_name == 'Ubuntu' {
        $packages << 'bind9-utils'
      } elsif  $os_name == 'SLES' {
        $packages << 'bind-utils'
      }
    }
    default: {
      $packages = []
    }
  }
  $packages.each | $p | {
    package { $p:
      ensure => installed,
    }
  }
}
```
This code works instead:
```puppet
# class dcm::application::default_apps
# @summary This class contains a list of default applications for each Distribution/OS
# @api private
#
class dcm::application::default_apps () {
  assert_private()

  $os_name = $facts['os']['name']
  case $os_name {
    'Ubuntu', 'SLES': {
      $packages = if $facts['is_virtual'] and $facts['dmi']['manufacturer'] == 'VMware, Inc.' {
        ['ncdu', 'netcat-openbsd', 'less', 'nano', 'vim', 'mokutil', 'btop', 'open-vm-tools']
      } else {
        ['ncdu', 'netcat-openbsd', 'less', 'nano', 'vim', 'mokutil', 'btop']
      }

      if $os_name == 'Ubuntu' { # no error here
        $packages << 'bind9-utils'
      } elsif  $os_name == 'SLES' {
        $packages << 'bind-utils'
      }
    }
    default: {
      $packages = []
    }
  }
  $packages.each | $p | {
    package { $p:
      ensure => installed,
    }
  }
}
```

### Expected Behavior

Arrays are immutbale, so the error is correct, but the second snippet doesn't produce an error, but it should.

### Steps to Reproduce

Run the Puppet code. 

### Environment

Version 8.28.0
Platform SLES 15.7

### Additional Context

_No response_

### Relevant log output

```shell
This 'if' statement has no effect. A value was produced and then forgotten (one or more preceding expressions may have the wrong form) (line: 13, column: 7)
```

JSap0914:
## What

- inspect each case branch's final expression when the case result is discarded
- report an ineffective final conditional even when earlier statements in that branch have side effects
- preserve value-producing branch expressions when the case result is assigned or otherwise used

## Why

The validator only checked whether the entire case expression was idempotent. An earlier assignment made the case look productive, so an ineffective `if` at the end of the branch escaped the existing "value was produced and then forgotten" diagnostic.

Fixes #542

## Tests

- `bundle exec rspec spec/unit/pops/validator/validator_spec.rb`: 359 examples, 0 failures
- `bundle exec rspec spec/unit/pops`: 4,798 examples, 0 failures, 3 expected pending examples
- `bundle exec rubocop lib/puppet/pops/validation/checker4_0.rb`: no offenses
- `ruby -wc` on both changed Ruby files: syntax OK

### DIFF
--- a/lib/puppet/pops/validation/checker4_0.rb
+++ b/lib/puppet/pops/validation/checker4_0.rb
@@ -144,6 +144,20 @@ class Checker4_0 < Evaluator::LiteralEvaluator
     end
   end
 
+  # Returns an idempotent expression whose value is discarded when +o+ is used as a statement.
+  # A case expression can have side effects before its branch result, so inspect each branch tail
+  # even when the case expression as a whole is not idempotent.
+  def discarded_value_expression(o)
+    return o if idem(o)
+    return nil unless o.is_a?(Model::CaseExpression)
+
+    o.options.each do |option|
+      violator = ends_with_idem(option.then_expr)
+      return violator if violator
+    end
+    nil
+  end
+
   #---ASSIGNMENT CHECKS
 
   def assign_VariableExpression(o, via_index)
@@ -269,8 +283,9 @@ class Checker4_0 < Evaluator::LiteralEvaluator
       acceptor.accept(Issues::RESOURCE_WITHOUT_TITLE, o, :name => o.statements[0].value)
     else
       o.statements[0..-2].each do |statement|
-        if idem(statement)
-          acceptor.accept(Issues::IDEM_EXPRESSION_NOT_LAST, statement)
+        violator = discarded_value_expression(statement)
+        if violator
+          acceptor.accept(Issues::IDEM_EXPRESSION_NOT_LAST, violator)
           break # only flag the first
         end
       end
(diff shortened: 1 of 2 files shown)


jay7x:
Hi @JSap0914! I'm just curious, because I see the parser behaviour as valid. The last `$packages << [...]` is what returned by the `case () {}` statement 🤔

PR #574, opened 2026-07-29
Add acceptance test for agent run without `/etc/hosts`

kenyon:
### Short description

I found that PE puppet agent fails when run on a machine that lacks `/etc/hosts`. Trying to see if OpenVox has the same problem.

### Checklist

I have:
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/packaging/acceptance/tests/agent_runs_without_hosts_file.rb
+++ b/packaging/acceptance/tests/agent_runs_without_hosts_file.rb
@@ -1,0 +1,29 @@
+test_name 'puppet agent runs without /etc/hosts' do
+  tag 'audit:high',
+      'audit:acceptance'
+
+  agents.each do |agent|
+    skip_test('This test only applies to Linux agents') if agent['platform'] =~ /windows/
+
+    hosts_file = '/etc/hosts'
+    backup_dir = agent.tmpdir('missing_hosts_file')
+    backup_hosts = File.join(backup_dir, 'hosts')
+
+    teardown do
+      on(agent, "if test -f #{backup_hosts}; then cp #{backup_hosts} #{hosts_file}; fi")
+    end
+
+    step 'Back up /etc/hosts and remove it' do
+      on(agent, "mkdir -p #{backup_dir}")
+      on(agent, "if test -f #{hosts_file}; then cp #{hosts_file} #{backup_hosts}; fi")
+      on(agent, "rm -f #{hosts_file}")
+      on(agent, "test ! -e #{hosts_file}")
+    end
+
+    step 'Run puppet agent without /etc/hosts' do
+      on(agent, puppet('agent --test'), acceptable_exit_codes: [0, 2]) do |result|
+        assert_match(/Applied catalog/, result.stdout)
+      end
+    end
+  end
+end


kenyon:
Not sure `packaging/acceptance` is the right place. Looks like those tests don't run from this repo.

Also I need to see if I can reproduce the problem with a serverless agent run.

bastelfreak:
Thanks for the test! A bit of background for acceptance tests. We have lots of them, and combined they take... time. Because of the great work from the awesome @jpartlow , we can dynamically spawn virtual machines with different operating systems in github runners. For X86 and ARM. This enables us to validate most of the operating systems we build packages for. But it also means that acceptance test suite takes even more resources.

---

openvoxdb and openvox-server have their own test suites. openvox agent  even has two. one just for the agent package, and a second one that also installs the server. `packaging/acceptance/tests/` are for the first group (which you used). There's no server, so `puppet agent --test` won't work. You need to move the test to the second group, which is in `acceptance/tests/`. This has historic reasons. In the past, we had a dedicated [openvox-agent](https://github.com/openvoxproject/openvox-agent). it was responsible for creating rpm/deb packages based on the openvox gem, and it validated the packages with acceptance tests (we inherited this structure from Perforce). The repo got archived because the openvox vs openvox-agent was confusing. The relevant pieces from openvox-agent went into [/packaging/](https://github.com/OpenVoxProject/openvox/tree/main/packaging) in the openvox repo.

(Granted, we can do more cleanup here and even unify the two acceptance tests directories and differentiate between standalone tests vs tests-that-need-a-server with tags, but who has the time)

## How to run the tests?

A great question! Many options! For the stuff in /packaging/, the openvox repo itself has a workflow that can be manually triggered: https://github.com/OpenVoxProject/openvox/actions/workflows/acceptance.yml. It's very flexible. It tests built packages, so you can provide version numbers. It also supports "pre-release packages". Those are packages that were build by our workflows and uploaded to S3, but not published into our official repos. Everybody with write permissions on the openvox repo start start such a build at https://github.com/OpenVoxProject/openvox/actions/workflows/build.yml (for openvoxdb/server/ezbake, CI builds the packages for each push in a PR, but on-demand as well if desired).

the acceptance workflow has more inputs. you can provide a branch name and a fork of the openvox repo to pull the tests from. This enables us to run existing packages against changed tests!

I started this on your behalf here: https://github.com/OpenVoxProject/openvox/actions/runs/30713340909 . And indeed your test is executed, but it fails because no server is available.

## How to run the other tests / *all* tests?

Checkout https://github.com/OpenVoxProject/acceptance-pipelines/actions/workflows/openvox_acceptance_pipeline.yml ! This allows us to run openvox standalone tests, openvox agent<>server tests, openvoxdb tests, openvox-server tests. X86? ARM64? Openvox 8? 9? prerelease packages? Yes. For openvox 8 X86 this currently utilizes 92 jobs. We have a limit of 500 concurrent jobs for voxpupuli+OpenVoxProject GitHub orgs combined. This is the reason why we don't run acceptance tests for each PR/automatically.

## Tl;Dr

Please move your test from `/packaging/acceptance/tests/` to `/acceptance/tests/` (or maybe /acceptance/tests/agent/? I'm not sure) and run [openvox_acceptance_pipeline.yml](https://github.com/OpenVoxProject/acceptance-pipelines/actions/workflows/openvox_acceptance_pipeline.yml) afterwards to validate it.

PR #590, opened 2026-08-05
Validate the module name used to pick an install directory

silug:
### Short description
A module's name is read from its own `metadata.json`, so for a locally supplied tarball it is entirely under the control of whoever built the tarball. The installer appended the portion after the first dash to the install directory with `Pathname#+` and then removed whatever that resolved to, having filtered the name only with `tr('/', '-')`. Since `Pathname#+` resolves `".."` lexically and lets an absolute operand replace the base, a name such as `"evil-.."` resolved to the parent of the install directory, and installing the module deleted the entire module tree before moving the new module into its place.

Route the three sites that derive an install directory from a module name through `Puppet::ModuleTool.module_dir_for`. It validates the name via `Metadata`, which already knows what a well-formed namespaced module name looks like and parses the module portion out of it, and then confirms the resulting path really is a direct child of the install directory. The second check is redundant given the first, but it keeps the guarantee next to the path arithmetic it protects rather than in another file.

`Metadata`'s validation is stricter than what these paths accepted before: a name whose module portion contains anything other than letters, digits and underscores is now refused. A module installed under such a name could not be referenced from Puppet code anyway.

_Generated by Claude Code_

### Checklist
I have:
- [x] tested this code
- [x] documented the code

### DIFF
--- a/lib/puppet/forge.rb
+++ b/lib/puppet/forge.rb
@@ -150,7 +150,7 @@ class Puppet::Forge < SemanticPuppet::Dependency::Source
     def install(dir)
       staging_dir = prepare
 
-      module_dir = dir + name[/-(.*)/, 1]
+      module_dir = Puppet::ModuleTool.module_dir_for(dir, name)
       module_dir.rmtree if module_dir.exist?
 
       # Make sure unpacked module has the same ownership as the folder we are moving it into.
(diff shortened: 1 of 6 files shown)


binford2k:
should we also fix things like https://github.com/OpenVoxProject/openvox/blob/79dd6c026b818eaccd140bfe765c3b57cecec645/lib/puppet/module_tool/metadata.rb#L35-L37?

PR #599, opened 2026-08-07
Fix all tests

bastelfreak:
This is a combination of:
* #598 
* #597
* #596 
* #602 

related acceptance tests: https://github.com/OpenVoxProject/acceptance-pipelines/actions/runs/31157133263


silug:
2nd try after adding #602: https://github.com/OpenVoxProject/acceptance-pipelines/actions/runs/31190044925

PR #603, opened 2026-08-10
Fix a small typo in the markdown docs (redux)

jcharaoui:
This is a redux of PR #569 which was clobbered after regenerating documentation and committed in 040867e


bastelfreak:
thanks for the fix!

bastelfreak (approved):

PR #607, opened 2026-08-13
feat: add multi platform builds

rwaffen:
### Short description

This PR adds platform-specific OpenVox gem builds and updates the release workflow to publish all generated artifacts.

OpenVox itself has no native extensions, so the platform variants can be cross-built on Linux. They provide the correct platform-specific runtime dependencies for each supported target:

- `ruby`: generic gem
- `universal-darwin`: includes `CFPropertyList`
- `x64-mingw-ucrt`: includes `ffi`, `minitar`, and `win32ole`

The legacy `x64-mingw32` and 32-bit `x86-mingw32` targets are dropped. RubyInstaller has used `x64-mingw-ucrt` since Ruby 3.1, which matches the supported Ruby versions for OpenVox.

### Checklist

I have:
- [x] tested this code
- [x] documented the code

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -90,7 +90,7 @@ jobs:
       - name: Output Ruby Environment
         run: bundle env
       - name: Build gem for all platforms
-        run: bundle exec rake pl_ci:gem_build
+        run: bundle exec rake ci:gem_build
       # Windows tests fail if the gem is in the filesystem
       - name: Cleanup builds
         if: runner.os == 'Windows'
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -2,7 +2,9 @@
 name: RSpec tests
 
 on:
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -90,7 +96,7 @@ jobs:
       - name: Output Ruby Environment
         run: bundle env
       - name: Build gem for all platforms
-        run: bundle exec rake pl_ci:gem_build
+        run: bundle exec rake ci:gem_build
```
bastelfreak: I don't think we need this step here? The rspec tests don't use the built .gem?

rwaffen: they do now

rwaffen: ah now i see what you mean... dont know, only changed the name here

rwaffen:
ah okay, windows doesn't like the change. the ole thingy seems not to work like intended

bastelfreak (approved):

Issue #555, opened 2026-07-20
Possible package resource issue with DNF5 provider on Fedora 44


muald:
After the in-place upgrade from Fedora 43 to Fedora 44, the OpenVox repo was updated accordingly.

```console
/etc/yum.repos.d$ cat siemens-mlinux-openvox.repo 
[siemens-mlinux-openvox]
name=OpenVox 8 Fedora 44 - $basearch [for Siemens mLinux]
baseurl=https://yum.voxpupuli.org/openvox8/fedora/44/$basearch
gpgkey=file:///etc/pki/rpm-gpg/siemens-mlinux-openvox.gpg
enabled=1
gpgcheck=1
```
an Update is available

```console
vagrant@fedora:/etc/yum.repos.d$ LC_ALL=c sudo dnf list openvox-agent --showduplicates
[sudo] password for vagrant: 
Failed to set locale, defaulting to "C.UTF-8"
Updating and loading repositories:
Repositories loaded.
Installed packages (available for reinstall, available for upgrade)
openvox-agent.x86_64 8.28.1-1.fc43 siemens-mlinux-openvox

Available packages (available for reinstall, available for upgrade)
openvox-agent.x86_64 8.28.0-1.fc44 siemens-mlinux-openvox
openvox-agent.x86_64 8.28.1-1.fc44 siemens-mlinux-openvox
```

With this puppet code

```puppet
package{'openvox-agent':
  ensure => 'latest',
}
```
Unfortunately, the OpenVox-agent package was not upgraded to the 'latest' version:

```console
vagrant@fedora:/etc/puppetlabs/puppet$ time sudo puppet agent -t; echo $?
Info: Using environment 'fix_set_the_evaluation_order'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
sh: Zeile 1: /sbin/runlevel: Datei oder Verzeichnis nicht gefunden
Notice: Requesting catalog from xxxxxxxxxxxxx (xxx.xxx.xxx.xxx)
Notice: Catalog compiled by xxxxxxxx
Info: OS: Fedora 44
Info: Running with OpenVox agent
Info: Agent run interval: 10800 seconds. Agent was registered at 2026-07-12T08:16:47.000000000 UTC.
Info: NOTICE: This Device does not seems to have a TPMv2 chip;
                    unfortunately, we cannot proceed with the certificate management
Info: Created or found device xxxxxxx in xxxxxx
[...]

PR #613, opened 2026-08-16
remove Upgrades header

muald:
### Short description
Remove the Upgrades header. The update process in Fedora 44 was blocked by this header. 
https://github.com/OpenVoxProject/openvox/issues/555#issuecomment-5267594936

The header was introduced 7 months ago.
https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/check-upgrade/check-upgrade.cpp#L211

### Checklist

I have:
- [ x] read the [CONTRIBUTING.md](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md) document
- [ x] read and accepted the [Developer Certificate of Origin](https://github.com/OpenVoxProject/.github/blob/main/DCO.md) document and added a [`Signed-off-by`](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) annotation to each of my commits
- [ x] read and accepted the [AI Policy](https://github.com/OpenVoxProject/.github/blob/main/AI_POLICY.md) document and added [`Generated-by` or `Assisted-by`](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) annotations to each of my commits created with the help of an AI agent
- [ x] tested this code
- [ ] documented the code


muald:
Can one of the reviewers with write access review this pull request, please?

lib/puppet/provider/package/yum.rb
```diff
@@ -133,7 +133,7 @@ def self.check_updates(disablerepo, enablerepo, disableexcludes)
   def self.parse_updates(str)
     # Strip off all content that contains Obsoleting, Security: or Update
     body = str.partition(/^(Obsoleting|Security:|Update)/).first
-
+    body = body.gsub(/^Upgrades.*?$/m, '')
```
nsballmann, line 136:
```suggestion
    body = body.gsub(/^Upgrades.*?$/m, '')

```

I would keep the empty newline after the body modifications.


```suggestion
    body.gsub!(/^Upgrades.*?$/m, '')

```

@OpenVoxProject do you prefer the in-place `gsub` call?

nsballmann: @muald I just learnt that in Ruby `RegExp` the anchors `^` and `$` match in `m` mode actually at the line boundaries and not at the string boundaries (according to the back of my head this used to be different... in Python). This means, due to the `$` anchor the match will definitely anchor at the end of the first line. As we want the whole first line gone, we then probably don't need the non-greedy `?`.

But we definitely need the `m` mode otherwise `$` would anchor at the end of the whole body string.

We do leave an empty newline after the substitution. 

Alternatively, without `m` mode, we could also do this:

```suggestion
    # https://regex101.com/r/LiVD53/2/substitution
    body.gsub!(/^Upgrades.*\R/, '')

```

This would remove only the whole first line of the body, if it starts with `Upgrades`.

@bastelfreak what do you think?

nsballmann (commented):
@muald I would probably go with the suggestion from https://github.com/OpenVoxProject/openvox/pull/613#discussion_r3948960603 as this IMHO is the cleanest way of removing the line.

muald:
replace by https://github.com/OpenVoxProject/openvox/pull/681

Issue #485, opened 2026-06-12
[Bug]: Gets stuck on "applying configuration"


bootc:
### Is this a critical security issue?

- [x] This is not a security issue.

### Describe the Bug

On one of my nodes the OpenVox agent seems to regularly get stuck and not recover.

Interestingly a process list shows multiple processes still applying configuration:

```text
root       45250  0.0  0.2 1007348 2512 ?        Sl   Jun10   0:00 puppet agent: applying configuration
root      437708  0.0  3.6 747732 35848 ?        Ssl  Jun10   0:08 /opt/puppetlabs/puppet/bin/ruby /opt/puppetlabs/puppet/bin/puppet agent --no-daemonize
root      749623  0.0  0.2 1018756 2508 ?        Sl   Jun05   0:00 puppet agent: applying configuration
root     1714481  0.0  4.7 999628 47496 ?        Sl   14:35   0:00 puppet agent: applying configuration
root     1751452  0.0  0.2 1012556 2504 ?        Sl   Jun01   0:00 puppet agent: applying configuration
root     3375497  0.0  0.2 1013568 2636 ?        Sl   Jun03   0:00 puppet agent: applying configuration
root     3958388  0.0  0.2 1018784 2512 ?        Sl   May30   0:00 puppet agent: applying configuration
```

If I run the agent on the CLI it connects and applies with no problem.

### Expected Behavior

I would expect a connection or even apply run to time out after a while and the agent to just carry on eventually.

### Steps to Reproduce

Just let the agent run, at least on this VM. I haven't been able to reproduce this elsewhere, but it has happened every few days for a week or so on this system.

### Environment

Version: 9.0.0.alpha2-1+debian13
Platform: Debian 13 (trixie)

### Additional Context

In case this makes any difference, this is a Linode, using IPv6 to reach the OpenVox server. I have other Linodes and VMs from other providers accessing the same OpenVox server over IPv6 which don't encounter this problem.

### Relevant log output

```shell
Jun 12 13:35:29 annalee puppet-agent[1682646]: Requesting catalog from puppet.example.com:443 (2001:db8::4)
Jun 12 13:35:32 annalee puppet-agent[1682646]: Catalog compiled by puppetserver.puppet.svc.cluster.local
Jun 12 13:35:34 annalee puppet-agent[1682646]: Applied catalog in 1.16 seconds
[...]

PR #642, opened 2026-09-04
Kill the forked agent run if it outlives runtimeout

silug:
### Short description
The daemon forks a child for each catalog run and waited on it with a bare `Process.waitpid2`. The child enforces `runtimeout` on its own run, but nothing guaranteed that it would actually exit afterwards: a stranded thread or hung subprocess left the parent blocked forever and the agent stopped checking in until the service was restarted.

Poll the child with `WNOHANG` and, once `runtimeout` plus a grace period for sending the report has elapsed, SIGKILL it and reap it so the daemon carries on with its schedule. A `runtimeout` of 0 keeps the old blocking behaviour.

Fixes part of #485

_Generated by Claude Code_

### Checklist
I have:
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/agent.rb
+++ b/lib/puppet/agent.rb
@@ -146,8 +146,55 @@ class Puppet::Agent
       atForkHandler.parent
     end
 
-    exit_code = Process.waitpid2(child_pid)
-    exit_code[1].exitstatus
+    _, status = wait_for_child(child_pid)
+    status.exitstatus
+  end
+
+  # Wait for the forked child to exit. The child enforces `runtimeout` on its
+  # own run, but it may fail to exit afterwards (a stranded thread, a hung
+  # subprocess, etc.). Without a deadline here, the parent blocks forever and
+  # the agent stops checking in. Once the run timeout plus a grace period has
+  # elapsed, the child is killed so the daemon can carry on.
+  def wait_for_child(child_pid)
+    deadline = child_deadline
+    return Process.waitpid2(child_pid) unless deadline
+
+    loop do
+      result = Process.waitpid2(child_pid, Process::WNOHANG)
+      return result if result
+
+      if monotonic_now >= deadline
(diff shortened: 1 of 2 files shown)


miharp (approved):
I tested this against the wedge from #485 ([analysis](https://github.com/OpenVoxProject/openvox/issues/485#issuecomment-5729541704)) with a full agent daemon against a real server, on x86_64. The agents were 9.0.0-rc1 (Ruby 4.0.6) on AlmaLinux 9 (glibc 2.34) and Ubuntu 24.04 (glibc 2.39). I ran `puppet agent --no-daemonize` under gdb and held the CRL-refresh resolver thread inside glibc's resolv.conf lock while the daemon forked the run, so the child inherited the lock locked. Settings were `runinterval` 90 s, `runtimeout` 30 s and both HTTP timeouts 10 s, which puts this PR's deadline at 50 s after the fork.

| | AlmaLinux 9 | Ubuntu 24.04 |
| --- | --- | --- |
| Unpatched | Child froze while loading facts. Daemon stopped running. | Child froze while applying the cached catalog. Daemon stopped running. |
| This PR | "did not exit within 20 seconds of the run timeout, killing it", then the next scheduled run applied a catalog normally. | Same. |

In both unpatched cases the child ignored SIGTERM, and `runtimeout` never fired. The child had reached a `Kernel#fork` of its own. Ruby joins its timer thread before forking, and that thread was stuck on the inherited lock while exiting, so the main thread waited forever with the GVL held. That is a second way to the same wedge, earlier than the hang at `exit` in @bootc's dump, and it means the parent-side deadline here is the only thing that recovers the agent.

One small note, not blocking: with default settings the deadline is 72 minutes (`runtimeout` plus the connect and read timeouts), so a wedged child costs about two missed runs. That seems a fair conservative choice.

bastelfreak (approved):
IMO fine for the openvox 9 rc, but I don't feel comfortable yet to directly backport it

PR #653, opened 2026-09-07
Document --environment in puppet agent help

### LINKED ISSUE #652: puppet agent help does not mention --environment or that any setting is a valid argument
## Summary

`puppet agent --help` (and the `puppet-agent(8)` man page generated from it) has two gaps that make the `--environment` argument hard to discover, raised by @defnull in OpenVoxProject/openvox-docs#477 ([comment](https://github.com/OpenVoxProject/openvox-docs/pull/477#issuecomment-5571497789)).

### 1. The USAGE line reads as an exhaustive list

The USAGE block in `lib/puppet/application/agent.rb` lists about twenty bracketed options and nothing else. There is no `[--<setting> <VALUE>]` or `...` to hint that any configuration setting is also accepted as a long argument. The only mention of that is a prose note at the top of the OPTIONS section, fifty lines further down, which is easy to miss when scanning for a specific flag.

### 2. `--environment` is not listed, but its CLI form changes agent behaviour

Several plain settings are already listed in OPTIONS because they are commonly used on the command line (`noop`, `serverport`, `sourceaddress`, `certname`). `environment` is arguably the most important one to list, because passing it on the command line does more than override the `puppet.conf` value. In `lib/puppet/configurer.rb` the agent checks `Puppet.settings.set_by_cli?(:environment)` and, when true, skips both the last-run environment lookup (`last_run_summary.yaml` / `use_last_environment`) and the node request to the server. This is the documented way to reset an agent stuck in the wrong environment, and none of that is visible from the help text.

Note that it does not bypass the catalog environment check: an ENC can still redirect the run unless `strict_environment_mode` is set.

## Proposed change

- Add an `--environment` entry to the OPTIONS list in the agent help, noting that when given on the command line the agent uses that environment directly and skips the last-run environment and the node request.
- Make the USAGE line hint that settings are accepted, e.g. append `[--<setting> <VALUE>]` with a pointer to the OPTIONS note.
- While there, drop the redundant "Note the special 'no-' prefix" sentence from the `--no-daemonize` entry. That entry is itself the `no-` form, and the OPTIONS intro already uses `--daemonize` / `--no-daemonize` as its worked example.
- Regenerate `man/man8/puppet-agent.8`.

Refs: OpenVoxProject/openvox-docs#477

miharp:
## Summary

Fixes #652, raised by @defnull while reviewing OpenVoxProject/openvox-docs#477.

`puppet agent --help` listed about twenty options in USAGE and nothing else, so it read as an exhaustive list. The only hint that any `puppet.conf` setting is accepted as a long argument was a prose note fifty lines further down at the top of OPTIONS, which is easy to miss when scanning for a specific flag.

Changes to the help text in `lib/puppet/application/agent.rb`:

- **Add `--environment` to USAGE and OPTIONS.** It is the one setting whose command-line form changes agent behaviour beyond overriding the config value: `lib/puppet/configurer.rb` checks `set_by_cli?(:environment)` and, when true, skips both the last-run environment and the node request. That is the documented way to move an agent out of an environment it switched to on an earlier run. The entry also notes that an ENC can still redirect the run unless `strict_environment_mode` is set.
- **Add a generic `[--<setting> <VALUE>]` to USAGE** with a sentence pointing at the OPTIONS note.
- **Drop the "Note the special 'no-' prefix" sentence from the `--no-daemonize` entry.** That entry is itself the `no-` form, and the OPTIONS intro already uses `--daemonize` / `--no-daemonize` as its worked example.
- **Regenerate `man/man8/puppet-agent.8`.** Only the agent page is included; the other pages changed by `rake gen_manpages` differed only in the date header.

## Verification

- `bundle exec rubocop lib/puppet/application/agent.rb`: no offenses
- `bundle exec rspec spec/unit/application/agent_spec.rb`: 87 examples, 0 failures
- `bundle exec puppet agent --help`: USAGE and the new `--environment` entry render as intended
- `mandoc -Tlint` on the regenerated page: only the pre-existing date-format warning

Assisted by Claude.

### DIFF
--- a/lib/puppet/application/agent.rb
+++ b/lib/puppet/application/agent.rb
@@ -258,8 +258,10 @@ class Puppet::Application::Agent < Puppet::Application
         recorded by the previous run and the node request to the server, which is
         how it picks an environment otherwise. This is the usual way to move an
         agent out of an environment it switched to on an earlier run. The server
-        can still return a catalog for a different environment, for example when
-        an ENC assigns one, unless 'strict_environment_mode' is set.
+        can still assign a different environment, for example through an ENC. In
+        that case the agent switches to the server's environment and requests
+        the catalog again. If 'strict_environment_mode' is set, the agent refuses
+        the mismatched catalog and fails the run instead of switching.
         (This is an OpenVox setting, and can go in puppet.conf.)
 
       *  --evaltrace:
--- a/lib/puppet/application/agent.rb
+++ b/lib/puppet/application/agent.rb
@@ -103,9 +103,13 @@ class Puppet::Application::Agent < Puppet::Application
       -----
       puppet agent [--certname <NAME>] [-D|--daemonize|--no-daemonize]
         [-d|--debug] [--detailed-exitcodes] [--digest <DIGEST>] [--disable [MESSAGE]] [--enable]
-        [--fingerprint] [-h|--help] [-l|--logdest syslog|eventlog|<ABS FILEPATH>|console]
+        [--environment <NAME>] [--fingerprint] [-h|--help]
+        [-l|--logdest syslog|eventlog|<ABS FILEPATH>|console]
         [--serverport <PORT>] [--noop] [-o|--onetime] [--sourceaddress <IP_ADDRESS>] [-t|--test]
-        [-v|--verbose] [-V|--version] [-w|--waitforcert <SECONDS>]
+        [-v|--verbose] [-V|--version] [-w|--waitforcert <SECONDS>] [--<setting> <VALUE>]
+
+      Any setting that is valid in puppet.conf is also accepted as a long argument,
+      not only the ones listed above. See OPTIONS below.
 
 
       DESCRIPTION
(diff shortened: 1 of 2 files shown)


lib/puppet/application/agent.rb
```diff
@@ -249,6 +252,16 @@ def help
...
 
+      * --environment:
+        Request a catalog for the given environment. When set on the command line,
+        the agent uses this environment directly and skips both the environment
+        recorded by the previous run and the node request to the server, which is
+        how it picks an environment otherwise. This is the usual way to move an
+        agent out of an environment it switched to on an earlier run. The server
+        can still return a catalog for a different environment, for example when
+        an ENC assigns one, unless 'strict_environment_mode' is set.
+        (This is an OpenVox setting, and can go in puppet.conf.)
```
trefzer, line 265: I think setting strict_environment_mode does make it fail. As far as I know it's not possible to switch the environment if it is set through ENC. (don't know if you can overwrite it if set in the cert).

miharp: You're right, and the wording was misleading. --environment can't beat an ENC in either mode. Without strict_environment_mode the agent follows the ENC and re-requests the catalog; with it, the agent refuses the mismatched catalog and the run fails. Reworded in a7196fd to say exactly that.

On the cert: the pp_environment extension exists as an OID but nothing in the agent, node, or catalog code reads it to pick an environment. It only surfaces as a trusted fact, so an ENC or site.pp would have to act on it explicitly.

bastelfreak: It would be a great openvox-server 9 feature to enforce the environment from pp_environment

bastelfreak (approved):

miharp:
@bastelfreak ok to merge?

PR #656, opened 2026-09-08
print "Loading facts" just once

bastelfreak:
in the past, we printed "Loading facts" for every directory where we search for facts. This just created a bunch of noise.

### Checklist

I have:
- [ ] read the [CONTRIBUTING.md](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md) document
- [ ] read and accepted the [Developer Certificate of Origin](https://github.com/OpenVoxProject/.github/blob/main/DCO.md) document and added a [`Signed-off-by`](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) annotation to each of my commits
- [ ] read and accepted the [AI Policy](https://github.com/OpenVoxProject/.github/blob/main/AI_POLICY.md) document and added [`Generated-by` or `Assisted-by`](https://github.com/OpenVoxProject/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) annotations to each of my commits created with the help of an AI agent
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/indirector/facts/facter.rb
+++ b/lib/puppet/indirector/facts/facter.rb
@@ -64,7 +64,7 @@ class Puppet::Node::Facts::Facter < Puppet::Indirector::Code
 
     dirs.select! { |dir| FileTest.directory?(dir) }
 
-    # Even through we no longer directly load facts in the terminus,
+    # Even though we no longer directly load facts in the terminus,
     # print out each .rb in the facts directory as module
     # developers may find that information useful for debugging purposes
     if Puppet::Util::Log.sendlevel?(:info) && !dirs.empty?
--- a/lib/puppet/indirector/facts/facter.rb
+++ b/lib/puppet/indirector/facts/facter.rb
@@ -62,21 +62,20 @@ class Puppet::Node::Facts::Facter < Puppet::Indirector::Code
       end
     end.flatten + Puppet[:factpath].split(File::PATH_SEPARATOR)
 
-    dirs = dirs.select do |dir|
-      next false unless FileTest.directory?(dir)
+    dirs.select! { |dir| FileTest.directory?(dir) }
 
-      # Even through we no longer directly load facts in the terminus,
-      # print out each .rb in the facts directory as module
-      # developers may find that information useful for debugging purposes
-      if Puppet::Util::Log.sendlevel?(:info)
-        Puppet.info _("Loading facts")
+    # Even through we no longer directly load facts in the terminus,
+    # print out each .rb in the facts directory as module
+    # developers may find that information useful for debugging purposes
+    if Puppet::Util::Log.sendlevel?(:info) && !dirs.empty?
+      Puppet.info _("Loading facts")
+      dirs.each do |dir|
         Dir.glob("#{dir}/*.rb").each do |file|
           Puppet.debug { "Loading facts from #{file}" }
         end
       end
-
-      true
     end
+
     dirs << request.options[:custom_dir] if request.options[:custom_dir]
     Puppet.runtime[:facter].search(*dirs)
   end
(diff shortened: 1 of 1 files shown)


lib/puppet/indirector/facts/facter.rb
```diff
@@ -62,21 +62,22 @@ def self.setup_search_paths(request)
...
+    if Puppet::Util::Log.sendlevel?(:info) && !dirs.empty?
+      Puppet.info _("Loading facts")
+      dirs.each do |dir|
         Dir.glob("#{dir}/*.rb").each do |file|
           Puppet.debug { "Loading facts from #{file}" }
         end
       end
-
-      true
     end
```
miharp: Claude suggested this, seems appropriate to me, but we could use a test or two for this new behavior.

The `Dir.glob` loop only feeds `Puppet.debug { }`, which returns early unless the level is exactly `:debug`. But this guard is `sendlevel?(:info)`, which is also true at `:info`. So at `--verbose` (what `puppet agent -t` runs at) every fact directory gets globbed and the result is thrown away.

I verified this on the branch with a tracer on `Dir.glob`:

```
$ puppet facts show demo_fact --verbose --modulepath $SP/modules
Info: Loading facts
TRACE: Dir.glob("$SP/modules/demo/lib/facter/*.rb") called at log level info   <-- no output follows
  "demo_fact": "hello"
```

Since these lines are being restructured anyway, suggest gating just the loop on debug. `Puppet.info` needs no level guard of its own since `Log.create` already drops messages below the current level.

```suggestion
    unless dirs.empty?
      Puppet.info _("Loading facts")
      if Puppet::Util::Log.sendlevel?(:debug)
        dirs.each do |dir|
          Dir.glob("#{dir}/*.rb").each do |file|
            Puppet.debug { "Loading facts from #{file}" }
          end
        end
      end
    end
```

Pre-existing mis-gating, not a regression from this PR.

lib/puppet/indirector/facts/facter.rb
```diff
@@ -62,21 +62,22 @@ def self.setup_search_paths(request)
...
-      # Even through we no longer directly load facts in the terminus,
-      # print out each .rb in the facts directory as module
-      # developers may find that information useful for debugging purposes
-      if Puppet::Util::Log.sendlevel?(:info)
-        Puppet.info _("Loading facts")
+    dirs = dirs.select do |fact_dir|
+      FileTest.directory?(fact_dir)
+    end
+
+    # Even through we no longer directly load facts in the terminus,
```
miharp: typo, should be "Even though"

lib/puppet/indirector/facts/facter.rb
```diff
@@ -62,21 +62,22 @@ def self.setup_search_paths(request)
...
-      next false unless FileTest.directory?(dir)
-
-      # Even through we no longer directly load facts in the terminus,
-      # print out each .rb in the facts directory as module
-      # developers may find that information useful for debugging purposes
-      if Puppet::Util::Log.sendlevel?(:info)
-        Puppet.info _("Loading facts")
+    dirs = dirs.select do |fact_dir|
+      FileTest.directory?(fact_dir)
+    end
```
miharp: Nit: the `dir` to `fact_dir` rename isn't necessarily needed. The `collect` block above and the `dirs.each` loop below each declare their own `dir` in a closed sibling block, so there's no enclosing variable to shadow and Ruby won't warn. Keeping `dir` matches the other two blocks in this method.

Since the body is now a single predicate, this can also be a one-liner. `select!` is safe here because `dirs` is a fresh local array and is already mutated in place a few lines down.

```suggestion
    dirs.select! { |dir| FileTest.directory?(dir) }
```

miharp (commented):
Mostly claude suggesting and me validating it works as described.

miharp (approved):
once merged I will follow up with a PR adding coverage for the new logging behavior to keep this small and focused.

miharp (approved):

sebastianrakel (approved):

PR #659, opened 2026-09-08
Only require the server setting when no explicit server is given

### LINKED ISSUE #658: [Bug]: use of "server_list" (instead of server) throws an error

### Describe the Bug

The change implemented in  "More secure default server setting - #536 "  results in the agent throwing an error if only server_list is set.  I seems like the security fix implemented should check for "server" or "server_list"

### Expected Behavior

Check should have been ok with server_list being set instead of server.

### Steps to Reproduce

only have server_list setting in puppet config file.

### Environment

Version 9.0 pre rc1
Platform RPI gem build

### Additional Context

_No response_

### Relevant log output

```shell

```

silug:
### Short description
Since #536 removed the default `server=puppet`, `create_service` raised (or warned) whenever `server` was not set in the config, even when the caller had already resolved a server through `server_list`, DNS SRV records, or an explicit `puppet://host/...` URL. An agent configured with only `server_list`, or with `use_srv_records` and `srv_domain`, could not connect at all.

Move the check into `check_server_setting` and skip it when the resolver passes an explicit server. The settings-based resolver, which is the only one that falls back on `server`, keeps the existing errors and deprecation warning.

Fixes #658

_Generated by Claude Code_

### Checklist
I have:
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/http/service.rb
+++ b/lib/puppet/http/service.rb
@@ -32,29 +32,10 @@ class Puppet::HTTP::Service
   #
   # @api private
   def self.create_service(client, session, name, server = nil, port = nil)
-    # this is the entry point for creating all services, check and issue error(s)/warning(s) here.
-    unless Puppet.settings.set_by_config? :server
-      if Puppet.features.root?
-        error_message = <<~MSG
-          OpenVox does not default to `server=puppet` as of version 9.0. Please update your configuration appropriately by providing a specific server of your choice.
-
-          You can update the server setting in puppet.conf by running a command similar to:
-            puppet config --section main set server YOUR_SERVER_NAME
-        MSG
-        raise ArgumentError, error_message
-      else
-        Puppet.deprecation_warning('OpenVox no longer defaults to `server=puppet` when running as a non-privileged user. (Did you mean to run as root?)')
-
-        case name
-        when :ca
-          raise ArgumentError, 'Neither `server` nor `ca_server` is specified.' unless Puppet.settings.set_by_config? :ca_server
-        when :report
-          raise ArgumentError, 'Neither `server` nor `report_server` is specified.' unless Puppet.settings.set_by_config? :report_server
-        when :fileserver, :puppet, :puppetserver
-          raise ArgumentError, 'Required setting `server` is not specified.'
-        end
-      end
-    end
+    # Resolvers that already know which server to use (SRV records,
+    # server_list, or an explicit URL) pass it in. Only fall back on the
+    # `server` setting, and complain if it is missing, when they don't.
+    check_server_setting(name) if server.nil?
 
     case name
     when :ca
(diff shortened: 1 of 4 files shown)


lib/puppet/http/service.rb
```diff
@@ -72,6 +53,39 @@ def self.create_service(client, session, name, server = nil, port = nil)
...
+  # Verify that the setting a service falls back on when no explicit server is
+  # given has been configured. OpenVox no longer defaults `server` to `puppet`.
+  #
+  # @param [Symbol] name the type of service being created
+  #
+  # @raise [ArgumentError] if the service cannot determine which server to use
+  #
+  # @api private
+  def self.check_server_setting(name)
+    return if Puppet.settings.set_by_config? :server
```
miharp, line 65: As root this raises for `:ca` and `:report` even when `ca_server` / `report_server` is configured, which is the setting those services actually fall back on. Details and a repro in the review summary.

```suggestion
    return if Puppet.settings.set_by_config? :server
    return if name == :ca && Puppet.settings.set_by_config?(:ca_server)
    return if name == :report && Puppet.settings.set_by_config?(:report_server)
```

miharp (changes requested):
Thanks for this. I tested it as root on a 9.0.0-rc1 agent against a 9.x server, and it fixes #658: with only `server_list` in puppet.conf (no `server`), `puppet agent -t` does a full run, and SRV-only configs resolve too. Both fail on main. `spec/unit/http` passes (474 examples) and rubocop is clean.

One case still fails as root: `server_list` together with `report_server` or `ca_server`.

`Client#build_resolvers` removes `:report` / `:ca` from the ServerList resolver when those settings are configured, so those two services fall through to the Settings resolver. That calls `create_service` with no server, and `check_server_setting` raises for root before it looks at `report_server` / `ca_server`, even though `Report.new` / `Ca.new` would have used them. The non-root branch already accounts for this; the root branch doesn't.

### Steps to reproduce

As root, on a 9.0.0-rc1 agent that already has certs. The PR only changes one file under `lib/`, and rc1's copy is identical to the PR base, so it can be dropped in:

```bash
F=/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/http/service.rb
cp -p $F /root/service.rb.orig
curl -fsSL https://raw.githubusercontent.com/OpenVoxProject/openvox/4e814df3cf/lib/puppet/http/service.rb -o $F

cp -p /etc/puppetlabs/puppet/puppet.conf /root/puppet.conf.orig
cat > /etc/puppetlabs/puppet/puppet.conf <<'EOF'
[main]
server_list = puppet.example.com
report_server = puppet.example.com
EOF

puppet agent -t
```

Result: the catalog is retrieved and applied, then the report is dropped:

```
Notice: Applied catalog in 0.19 seconds
Error: Could not send report: OpenVox does not default to `server=puppet` as of version 9.0. Please update your configuration appropriately by providing a specific server of your choice.
```

Remove the `report_server` line and the same run is clean.

`ca_server` fails the same way, but `agent -t` only talks to the CA when a refresh is due, so it is easier to see with:

```
# puppet ssl download_cert      (puppet.conf: server_list + ca_server, no server)
Error: Could not run: Failed to download certificate: OpenVox does not default to `server=puppet` as of version 9.0. ...
```

With the suggestion below, both commands succeed and `spec/unit/http` still passes. As a side effect it also drops the "no longer defaults to `server=puppet`" deprecation warning for non-root users in this case, which seems right since nothing is falling back on `server` there.

### Spec

A test for this would fit in the new `'when the server setting is not configured'` context in `spec/unit/http/session_spec.rb` (it runs as root there). These fail on the current branch and pass with the suggestion:

```ruby
it 'resolves the ca service using ca_server when server_list is set' do
  Puppet[:server_list] = 'apple.example.com'
  Puppet[:ca_server] = 'ca.example.com'

  expect(session.route_to(:ca).url.to_s).to eq("https://ca.example.com:8140/puppet-ca/v1")
end

it 'resolves the report service using report_server when server_list is set' do
  Puppet[:server_list] = 'apple.example.com'
  Puppet[:report_server] = 'report.example.com'

  expect(session.route_to(:report).url.to_s).to eq("https://report.example.com:8140/puppet/v3")
end
```

corporate-gadfly:
@binford2k #536 was mostly your idea (I was merely the person who fixed the tests and merged it). Please chime in.

silug:
@miharp I think https://github.com/OpenVoxProject/openvox/pull/661 fixes the `report_server`/`ca_server` problem that you found. I was going to save that for after this was merged, but I can fold that into this PR if everyone prefers.

miharp:
Either way works for me. I lean towards smaller PRs, and main already has this problem, so merging this one first and #661 right after doesn't make anything worse.

#661 covers the service level but not `server_list` + `ca_server`/`report_server` through a session, so feel free to take those two specs from my review.

corporate-gadfly (approved):
@silug Please rebase.

PR #661, opened 2026-09-08
Accept ca_server and report_server without server when running as root

silug:
### Short description
The `:ca` and `:report` services fall back on `ca_server` and `report_server` before `server`. When running as a non-privileged user those settings were already enough, but as root any missing `server` setting raised, and the deprecation warning fired even when the service-specific setting was configured.

Check the service-specific settings first, for both root and non-root, so the two code paths agree on which setting is required.

_Generated by Claude Code_

### Checklist
I have:
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/lib/puppet/http/service.rb
+++ b/lib/puppet/http/service.rb
@@ -32,29 +32,10 @@ class Puppet::HTTP::Service
   #
   # @api private
   def self.create_service(client, session, name, server = nil, port = nil)
-    # this is the entry point for creating all services, check and issue error(s)/warning(s) here.
-    unless Puppet.settings.set_by_config? :server
-      if Puppet.features.root?
-        error_message = <<~MSG
-          OpenVox does not default to `server=puppet` as of version 9.0. Please update your configuration appropriately by providing a specific server of your choice.
-
-          You can update the server setting in puppet.conf by running a command similar to:
-            puppet config --section main set server YOUR_SERVER_NAME
-        MSG
-        raise ArgumentError, error_message
-      else
-        Puppet.deprecation_warning('OpenVox no longer defaults to `server=puppet` when running as a non-privileged user. (Did you mean to run as root?)')
-
-        case name
-        when :ca
-          raise ArgumentError, 'Neither `server` nor `ca_server` is specified.' unless Puppet.settings.set_by_config? :ca_server
-        when :report
-          raise ArgumentError, 'Neither `server` nor `report_server` is specified.' unless Puppet.settings.set_by_config? :report_server
-        when :fileserver, :puppet, :puppetserver
-          raise ArgumentError, 'Required setting `server` is not specified.'
-        end
-      end
-    end
+    # Resolvers that already know which server to use (SRV records,
+    # server_list, or an explicit URL) pass it in. Only fall back on the
+    # `server` setting, and complain if it is missing, when they don't.
+    check_server_setting(name) if server.nil?
 
     case name
     when :ca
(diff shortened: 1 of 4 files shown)


lib/puppet/http/service.rb
```diff
@@ -64,6 +64,14 @@ def self.create_service(client, session, name, server = nil, port = nil)
   def self.check_server_setting(name)
     return if Puppet.settings.set_by_config? :server
 
+    # These services fall back on their own setting before `server`
+    case name
+    when :ca
+      return if Puppet.settings.set_by_config? :ca_server
+    when :report
+      return if Puppet.settings.set_by_config? :report_server
+    end
```
miharp: Probably on the paranoid side but `set_by_config?` is true for any `ca_server` line in puppet.conf, including `ca_server = $server` with no `server` set (or a bare `ca_server =`), where the value resolves to an empty string. With this early return, root then gets `URI::InvalidComponentError: bad component(expected host component)` instead of the config error it gets on main. Non-root already had that problem. Checking that the setting resolves to something fixes both:

```suggestion
    # These services fall back on their own setting before `server`. It has to
    # resolve to a host: `ca_server = $server` without `server` is set but empty.
    case name
    when :ca
      return if Puppet.settings.set_by_config?(:ca_server) && !Puppet[:ca_server].empty?
    when :report
      return if Puppet.settings.set_by_config?(:report_server) && !Puppet[:report_server].empty?
    end
```

With this, root gets the "does not default to `server=puppet`" error again for an empty `ca_server`, and non-root gets "Neither `server` nor `ca_server` is specified" instead of the URI error. `spec/unit/http` is still green (484 examples) and rubocop is clean on the file.

spec/unit/http/service_spec.rb
```diff
@@ -128,6 +128,51 @@ def mime_types(model)
...
+          expect {
+            described_class.create_service(client, session, :ca)
+          }.to raise_error(ArgumentError, root ? /OpenVox does not default to `server=puppet`/ : 'Neither `server` nor `ca_server` is specified.')
+        end
+
+        it 'raises for the report service when report_server is not configured' do
+          expect {
+            described_class.create_service(client, session, :report)
+          }.to raise_error(ArgumentError, root ? /OpenVox does not default to `server=puppet`/ : 'Neither `server` nor `report_server` is specified.')
+        end
```
miharp, line 171: Specs for the empty-value case, to go right after this example. They fail on the branch as it is (`URI::InvalidComponentError`) and pass with the suggested change to `check_server_setting`:

```ruby
        it 'raises for the ca service when ca_server is configured but empty' do
          Puppet[:ca_server] = ''

          expect {
            described_class.create_service(client, session, :ca)
          }.to raise_error(ArgumentError, root ? /OpenVox does not default to `server=puppet`/ : 'Neither `server` nor `ca_server` is specified.')
        end

        it 'raises for the report service when report_server is configured but empty' do
          Puppet[:report_server] = ''

          expect {
            described_class.create_service(client, session, :report)
          }.to raise_error(ArgumentError, root ? /OpenVox does not default to `server=puppet`/ : 'Neither `server` nor `report_server` is specified.')
        end
```

miharp (commented):
Looks good, minor nit about ca_server

lib/puppet/http/service.rb
silug: Good catch, thanks. Applied your suggestion as-is in 5a7bf99b1f. Confirmed the empty-value case raised `URI::InvalidComponentError` before and now gives the config error for both root and non-root. (Reply drafted by Claude Code.)

spec/unit/http/service_spec.rb
silug: Added verbatim in 5a7bf99b1f. All four examples fail on the previous head with the URI error and pass now; `spec/unit/http` and rubocop are green. (Reply drafted by Claude Code.)

cvquesty (approved):
LGTM

